PR #137, opened 2017-01-24
Fit and finish and bump up version to rc1

jianghaolu:


jianghaolu:
Will keep beta5 for now.

PR #172, opened 2017-05-25
Use 0.13 of Okio

jianghaolu:


dadoonet:
Where this artifact is coming from?

I can see:

`com.squareup.okio:okio:1.13.0`

Is this is a specific repo I'm missing?

PR #176, opened 2017-06-29
[After 1.0.4] Enabling user to poll long running operations

anuchandy:
**What is changed in AzureClient?**

***Following methods are added.***

**beginPutOrPatchAsync**

```java
AzureClient.<T> Single<PollingState<T>> 
        beginPutOrPatchAsync(Observable<Response<ResponseBody>> observable, 
                             final Type resourceType)
```

Initiate a LRO PUT or PATCH operation and returns serializable PollingState representing the LRO. This state can be used later to resume the polling.

**beginPostOrDeleteAsync**

```java
AzureClient.<T> Single<PollingState<T>>  
        beginPostOrDeleteAsync(Observable<Response<ResponseBody>> observable, 
                               final Type resourceType)
```

Initiate a LRO POST or DELETE operation and returns serializable PollingState representing the LRO. This state can be used later to resume the polling.

**pollSingleAsync**

```java
AzureClient.<T> Single<PollingState<T>> 
        pollSingleAsync(final PollingState<T> pollingState, 
                        final Type resourceType)
```

Given a pollingState (state of PUT_PATCH or POST_DELETE) this method poll once and returns the current polling state

**pollAsync**

```java
AzureClient.<T> Observable<PollingState<T>> 
        pollAsync(final PollingState<T> pollingState, final Type resourceType)
```

Given a pollingState (of state of PUT_PATCH or POST_DELETE) this method keeps on polling and emits each pollingState. Since the return type is Observable, user can apply all operators to have further control (poll n times, timeout etc..)

***How this affects the auto generated code?***

The `begin[Action]Async` methods in the inner collection will eventually return `Single<PollingState<T>>` (i.e. they internally calls `beginPostOrDeleteAsync` or `beginPutOrPatchAsync`). Now user can use the above other methods in the AzureClient to do polling.

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/PollingState.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/PollingState.java
@@ -47,13 +47,13 @@ public class PollingState<T> {
     private int retryTimeout;
     /** The resource uri on which PUT or PATCH operation is applied. **/
     private String putOrPatchResourceUri;
-    /** The logging context **/
+    /** The logging context. **/
     private String loggingContext;
 
 
     // Non-serializable properties
     //
-    /** The logging context header name **/
+    /** The logging context header name. **/
     @JsonIgnore
     private static final String LOGGING_HEADER = "x-ms-logging-context";
     /** The Retrofit response object. */
@@ -85,6 +85,8 @@ public class PollingState<T> {
      * @param defaultRetryTimeout the long running operation retry timeout.
      * @param resourceType the type of the resource the long running operation returns
      * @param serializerAdapter the adapter for the Jackson object mapper
+     * @param <T> the result type
+     * @return the polling state
      * @throws IOException thrown by deserialization
      */
     public static <T> PollingState<T> create(Response<ResponseBody> response, int defaultRetryTimeout, Type resourceType, SerializerAdapter<?> serializerAdapter) throws IOException {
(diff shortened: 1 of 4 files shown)


anuchandy:
@jianghaolu I think we could make few more improvements in the AzureClient, giving it a try so please don't merge now.

anuchandy:
@jianghaolu it seems the simplification i had in mind is not possible, please continue review.

azure-client-runtime/src/main/java/com/microsoft/azure/PollingState.java
```diff
@@ -141,28 +202,50 @@ void updateFromResponseOnDeletePost(Response<ResponseBody> response) throws IOEx
...
+        return this.putOrPatchResourceUri;
     }
 
     /**
-     * @return the resource type
+     * @return true if the status this state hold represents terminal status.
      */
-    Type resourceType() {
-        return resourceType;
+    boolean isStatusTerminal() {
```
jianghaolu: is this intended to be package private?

anuchandy: yes, thats what i thought while implementing theses methods [isStatusTerminal, isStatusFailed, isStatusSucceeded]. We expose 'status' to user. Do you think it is useful for user to expose these methods?

jianghaolu: No.. package private is safe to start with :)

PR #182, opened 2017-08-02
Add ExpandableStringEnum

daschult:


jianghaolu:
Do you think this should belong in `client-runtime` instead?

PR #184, opened 2017-08-04
Add json type info to expandable string enum

daschult:
Without my change to Validator.java, one of our tests was throwing a StackOverflowException because the Validator.validate() function would attempt to validate a Class type, then it would attempt to validate a Constructor type, and then it would go back to a Class type. This stops that cycle because we don't need to validate language level classes.


jianghaolu:
https://travis-ci.org/Azure/autorest-clientruntime-for-java/builds/261930095#L2453

jianghaolu:
@anuchandy Could you review the changes to lro?

azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java
```diff
@@ -400,7 +400,8 @@ public Boolean call(PollingState<T> tPollingState) {
         pollingState.withResourceType(resourceType);
         pollingState.withSerializerAdapter(restClient().serializerAdapter());
         if (pollingState.isStatusTerminal()) {
-            if (pollingState.isStatusSucceeded()
+            if (pollingState.statusCode() != 204
```
anuchandy: LGTM, @jianghaolu improved it by adding one more check so that we don't make a call if the resource type is Void.

anuchandy:
LGTM. Seems i cannot approve, may be due to the conflict in AzureClient.

daschult:
@anuchandy @jianghaolu Everything seems to pass. I'd appreciate it if someone could merge this please.

jianghaolu:
Not yet. Not for 1.1

daschult:
@jianghaolu Is this pull request okay to merge now?

PR #185, opened 2017-08-06
[For discussion]: HTTP annotation interfaces 

anuchandy:


client-runtime/src/main/java/com/microsoft/rest/v2/annotations/EndpointHost.java
```diff
@@ -0,0 +1,58 @@
...
+ * The applicability of this annotation is limited to variable with name {host} in the URI template.
+ *
+ *   When this annotation is applied to a method or interface level, value must be defined statically.
+ *   When this annotation is applied to a method parameter level, the parameter value will be it's value
+ *
+ * TODO: Decide whether we need all three of below or subset of it.
+ *
+ *  [1] Annotation applied to interface level.
+ *
+ *     @EndpointHost("manage.windowsazure.com")
```
jianghaolu: As we discussed this may not be necessary

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/EndpointHost.java
```diff
@@ -0,0 +1,58 @@
...
+ *        @EndpointHost("manage.windowsazure.com")
+ *        @GET("https://{host}/subscriptions/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
+ *        VirtualMachine getByResourceGroup(@PathParam("resourceGroupName") String rgName, @PathParam("vmName") String vmName, @PathParam("subscriptionId") String subscriptionId);
+ *    }
+ *
+ *  [3] Annotation applied to method parameter level.
+ *
+ *    interface VirtualMachines {
+ *        @GET("https://{host}/subscriptions/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
+ *        VirtualMachine getByResourceGroup(@EndpointHost String host, @PathParam("resourceGroupName") String rgName, @PathParam("vmName") String vmName, @PathParam("subscriptionId") String subscriptionId);
```
jianghaolu: A more common scenario in AutoRest is parametrized host - we need to treat host just like path and allow parameters

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/EndpointPort.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * The applicability of this annotation is limited to variable with name {protocol} in the URI template.
+ *   When this annotation is applied to a method or interface level, value must be defined statically.
+ *   When this annotation is applied to a method parameter level, the parameter value will be it's value
+ *
+ * TODO: Decide whether we need all three of below or subset of it.
+ *
+ *  [1] Annotation applied to interface level.
+ *
+ *        @EndpointPort(443)
```
jianghaolu: Same here

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/EndpointPort.java
```diff
@@ -0,0 +1,60 @@
...
+ *    interface VirtualMachines {
+ *        @EndpointPort(443)
+ *        @GET("https://manage.windowsazure.com:{port}/subscriptions/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
+ *        VirtualMachine getByResourceGroup(@PathParam("resourceGroupName") String rgName, @PathParam("vmName") String vmName, @PathParam("subscriptionId") String subscriptionId);
+ *    }
+ *
+ *  [3] Annotation applied to method parameter level.
+ *
+ *    interface VirtualMachines {
+ *        @GET("https://manage.windowsazure.com:{port}/subscriptions/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
```
jianghaolu: This can be part of the parametrized host

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/EndpointProtocol.java
```diff
@@ -0,0 +1,58 @@
...
+ *    interface VirtualMachines {
+ *        @EndpointProtocol("https")
+ *        @GET("{protocol}://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
+ *        VirtualMachine getByResourceGroup(@PathParam("resourceGroupName") String rgName, @PathParam("vmName") String vmName, @PathParam("subscriptionId") String subscriptionId);
+ *    }
+ *
+ *  [3] Annotation applied to method parameter level.
+ *
+ *    interface VirtualMachines {
+ *        @GET("{protocol}://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
```
jianghaolu: Same for this

jianghaolu: Maybe not for protocol -

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/PathParam.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ *  Example#3:
+ *
+ *    If the "uri template" is relative then engine will resolve full url using the base url. If base URL contains path segment variables then
+ *    that also should be resolved using the parameters annotated with PathParam.
+ *
+ *    @GET("/foo/{name2}/values")
+ *    String getValues(@PathParam("name1") String param1, @PathParam("name2") String param2);
+ *
+ *    Here name2 is not exists in the uri segment but after resolving the full path URI may looks like:
```
jianghaolu: This will make things a lot more complicated - we can't throw error during construction of the dynamic proxy - we have to wait until user provides input to see if it's valid or not. If this is requested in the future, it should be able to be added in a non-breaking way.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/EndpointHost.java
jianghaolu: Also this gets even trickier in the case of multiple environments - I'll send a PR to propose some of my thoughts

PR #186, opened 2017-08-07
V2 more

jianghaolu:


azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
```diff
@@ -0,0 +1,66 @@
...
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+
+@Target(value={METHOD,TYPE})    // The context in which annotation is applicable i.e. this annotation (Host) can be applied to method, method parameter and interface.
+@Retention(RetentionPolicy.RUNTIME)        // Record this annotation in the class file and make it available during runtime.
+public @interface AzureHost {
+    String value() default "";
+    AzureEnvironment.Endpoint enpoint() default Endpoint.RESOURCE_MANAGER;
```
jianghaolu: This is my proposal for multiple environments:

```
@AzureHost(Endpoint.KEY_VAULT)
@Get("secrets/{secretName}")
Secret get(@HostParam String vaultName, @PathParam("secretName") secretName);
```

daschult: Typo: endpoint(), not enpoint()

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
```diff
@@ -0,0 +1,66 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure.v2.annotations;
```
daschult: I'm a little concerned about putting version numbers into our packages/folder structure. Why not just create a v2 branch off of master that we do all of our work in until we're ready for v2 to be our main version, at which point we would just merge v2 into master and then delete v2?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
```diff
@@ -0,0 +1,66 @@
...
+
+@Target(value={METHOD,TYPE})    // The context in which annotation is applicable i.e. this annotation (Host) can be applied to method, method parameter and interface.
+@Retention(RetentionPolicy.RUNTIME)        // Record this annotation in the class file and make it available during runtime.
+public @interface AzureHost {
+    String value() default "";
+    AzureEnvironment.Endpoint enpoint() default Endpoint.RESOURCE_MANAGER;
+}
+
+/**
+ * The applicability of this annotation is limited to variable with name {host} in the URI template.
```
daschult: These comments are great, well written, and thought-out. Why not make them Javadoc comments on the @interface type itself so that they're available while we're writing code?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ *
+ *  [1] Static annotation applied to interface level.
+ *
+ *    @Host("manage.windowsazure.com")
```
daschult: Why is the interface named "AzureHost" but in the examples it uses "@Host"?

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -0,0 +1,44 @@
...
+    private RestProxy(String host, RestClient restClient) {
+        this.host = host;
+        this.restClient = restClient;
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+        if (method.isAnnotationPresent(GET.class)) {
+            String relativePath = method.getAnnotation(GET.class).value();
+            return (host + "/" + relativePath).getBytes();
```
daschult: Is this supposed to actually make the REST API network call at this point using RxNetty/OkHttp? I'm confused why the proxy is just returned the bytes of the GET URL.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Contexts.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Retention(RUNTIME)
+@Target(METHOD)
+public @interface Contexts {
```
daschult: What would be an example of a Contexts annotation?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/DELETE.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD})            // The context in which annotation is applicable i.e. this annotation (GET) can be applied only to methods
+@Retention(RetentionPolicy.RUNTIME)      // Record this annotation in the class file and make it available during runtime.
+public @interface DELETE {
+    String value() default "";
```
daschult: Would it be better to not provide a default value so that we don't forget to add the DELETE request's URL?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/DELETE.java
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ *  For more details on format of "value" field and associated rules refer {@link com.microsoft.rest.Doc_Http_Verb_Annotation_Value_URI_Template}
+ *
+ *  Below shows an example where 'DELETE' annotation is applied to the method with relative path uri template:
+ *
+ *  @GET("subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
```
daschult: Update example to use a DELETE annotation instead of GET.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Doc_Http_Verb_Annotation_Value_URI_Template.java
```diff
@@ -36,7 +36,7 @@
  *  The name of these parameters are predefined, if they are present then they must be present exactly in places where they are suppose to present in a valid URI.
  *
  *  Value of {protocol} is derived from {@link EndpointProtocol} annotation. Refer documentation of {@link EndpointProtocol} for more details.
- *  Value of {host} is derived from {@link EndpointHost} annotation. Refer documentation of {@link EndpointHost} for more details.
+ *  Value of {host} is derived from {@link Host} annotation. Refer documentation of {@link Host} for more details.
```
daschult: Why did you change this from EndpointHost to Host when the other two values have an "Endpoint" prefix?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
```diff
@@ -0,0 +1,66 @@
...
+import com.microsoft.azure.AzureEnvironment.Endpoint;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+
+@Target(value={METHOD,TYPE})    // The context in which annotation is applicable i.e. this annotation (Host) can be applied to method, method parameter and interface.
```
daschult: EndpointHost could be targeted to a PARAMETER. Is there a reason you don't allow AzureHost to target a PARAMETER?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/HEAD.java
```diff
@@ -0,0 +1,28 @@
...
+public @interface HEAD {
+    String value() default "";
+}
+
+/**
+ *  For more details on format of "value" field and associated rules refer {@link com.microsoft.rest.Doc_Http_Verb_Annotation_Value_URI_Template}
+ *
+ *  Below shows an example where 'HEAD' annotation is applied to the method with relative path uri template:
+ *
+ *  @GET("subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
```
daschult: Update example to use HEAD.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/HeaderParam.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Retention(RUNTIME)
+@Target(PARAMETER)
+public @interface HeaderParam {
```
daschult: Can you add an example for this annotation?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Headers.java
```diff
@@ -0,0 +1,22 @@
...
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Retention(RUNTIME)
+@Target(METHOD)
+public @interface Headers {
+    /**
+     * List of static headers.
+     */
+    String[] value();
```
daschult: Do these strings take the format "headerName=headerValue1,headerValue2"? Does that mean that the annotation parser will have to parse the list of headers?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
daschult: Typo: endpoint(), not enpoint()

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
daschult: Can you change these examples to use AzureHost instead of Host?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Host.java
```diff
@@ -0,0 +1,62 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+
+@Target(value={METHOD,TYPE})    // The context in which annotation is applicable i.e. this annotation (Host) can be applied to method, method parameter and interface.
+@Retention(RetentionPolicy.RUNTIME)        // Record this annotation in the class file and make it available during runtime.
+public @interface Host {
```
daschult: Why do we allow both AzureHost and Host? What is the benefit of using one over the other?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/PATCH.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.microsoft.rest.v2.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD})            // The context in which annotation is applicable i.e. this annotation (GET) can be applied only to methods
```
daschult: Change comment annotation reference from (GET) to (PATCH).

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/PATCH.java
```diff
@@ -0,0 +1,28 @@
...
+public @interface PATCH {
+    String value() default "";
+}
+
+/**
+ *  For more details on format of "value" field and associated rules refer {@link com.microsoft.rest.Doc_Http_Verb_Annotation_Value_URI_Template}
+ *
+ *  Below shows an example where 'PATCH' annotation is applied to the method with relative path uri template:
+ *
+ *  @GET("subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
```
daschult: Update example to use @PATCH annotation instead of @GET.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/POST.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.microsoft.rest.v2.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD})            // The context in which annotation is applicable i.e. this annotation (GET) can be applied only to methods
```
daschult: Update comment reference to use (POST) instead of (GET).

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/POST.java
```diff
@@ -0,0 +1,28 @@
...
+public @interface POST {
+    String value() default "";
+}
+
+/**
+ *  For more details on format of "value" field and associated rules refer {@link com.microsoft.rest.Doc_Http_Verb_Annotation_Value_URI_Template}
+ *
+ *  Below shows an example where 'POST' annotation is applied to the method with relative path uri template:
+ *
+ *  @GET("subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
```
daschult: Update example to use @POST instead of @GET.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/PUT.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.microsoft.rest.v2.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD})            // The context in which annotation is applicable i.e. this annotation (GET) can be applied only to methods
```
daschult: Update comment reference to be (PUT) instead of (GET).

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/PUT.java
```diff
@@ -0,0 +1,28 @@
...
+@Retention(RetentionPolicy.RUNTIME)      // Record this annotation in the class file and make it available during runtime.
+public @interface PUT {
+    String value() default "";
```
daschult: Each of the HTTP Method annotations have a field named "value", even though value is actually the URL path. Is there a reason we didn't name it "path"?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/PUT.java
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/**
+ *  For more details on format of "value" field and associated rules refer {@link com.microsoft.rest.Doc_Http_Verb_Annotation_Value_URI_Template}
+ *
+ *  Below shows an example where 'PUT' annotation is applied to the method with relative path uri template:
+ *
+ *  @GET("subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachines/{vmName}")
```
daschult: Update example to use PUT.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/QueryParam.java
```diff
@@ -0,0 +1,28 @@
...
+     * The name of the variable in the endpoint uri template which will be replaced with the value
+     * of the parameter annotated with this annotation.
+     */
+    String value();
+    /**
+     * A value true for this argument indicates that value of {@link QueryParam#value()} is already encoded
+     * hence engine should not encode it, by default value will be encoded.
+     */
+    boolean encoded() default false;
+}
```
daschult: Can you provide an example of using this annotation?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/QueryParam.java
```diff
@@ -0,0 +1,28 @@
...
+public @interface QueryParam {
+    /**
+     * The name of the variable in the endpoint uri template which will be replaced with the value
+     * of the parameter annotated with this annotation.
+     */
+    String value();
+    /**
+     * A value true for this argument indicates that value of {@link QueryParam#value()} is already encoded
+     * hence engine should not encode it, by default value will be encoded.
+     */
```
daschult: Consider this: "This value indicates whether or not this parameter's value is encoded. If the parameter's value is encoded, the engine will not attempt to encode it again. The default value is false, meaning the engine will attempt to encode the parameter's value."

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
jianghaolu: The convention in Java world is to add version numbers to package names so that multiple versions can be used together

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
jianghaolu: Because this is not supported by Swagger 2.0 - to hit a completely different endpoint (parametrized ones are allowed) you need a separate swagger doc anyways

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
jianghaolu: I think we started out here because these are not actually finalized javadocs? @anuchandy I agree they will go into javadocs largely

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
jianghaolu: This is a just a place holder - > this is where your work will do the magic :)

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Contexts.java
jianghaolu: logging context.
Or anything that you'd like to pass down the pipeline

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/DELETE.java
jianghaolu: I think this is the same for @GET which is allowing @Url parameters to specify the entire path

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Doc_Http_Verb_Annotation_Value_URI_Template.java
jianghaolu: I think the other 2 should go away

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Headers.java
jianghaolu: Yes and yes

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Host.java
jianghaolu: The `AzureEnvironment.Endpoint` enum isn't available in the base `client-runtime`

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
daschult: Fair enough.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: That's what I thought after I looked at the unit test at the end of the review.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Doc_Http_Verb_Annotation_Value_URI_Template.java
daschult: The prefix to the other two annotations should go away, or the other two annotations should go away?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/annotations/AzureHost.java
anuchandy: yes, i put it there just like that, once we finalize we can move them to javadoc section.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Contexts.java
anuchandy: Should this be a map instead of array of strings? so that we can easily lookup.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/DELETE.java
anuchandy: and if both `@url` and `value()` are empty, then it would be a get on the base url right? or an error?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/PUT.java
anuchandy: One reason i can think of is, the value can contain not just "path segments" but query params, fragments along with path segments.

daschult: That's a good point. In that case, I'd say we should have a Javadoc comment explaining what could go into value, or we should rename it to pathQueryFragment. (I prefer the Javadoc comment.)

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/Contexts.java
anuchandy: ahh ignore my comment, its statically defined.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/HeaderParam.java
```diff
@@ -12,6 +12,29 @@
...
+ *
+ * Example 1:
+ *   {@literal @}PUT("{functionId}")
+ *   Observable<Response<ResponseBody>> createOrReplace(@PathParam("functionId", encoded = true) String functionId, @BodyParam FunctionInner function, @HeaderParam("If-Match") String ifMatch);
+ *
+ *   "If-Match: user passed value" will show up as one of the headers.
+ *
+ * Example 2:
+ *   {@literal @}GET("subscriptions/{subscriptionId}/providers/Microsoft.ServiceBus/namespaces")
+ *   Observable<Response<ResponseBody>> list(@Path("subscriptionId") String subscriptionId, @Header("accept-language") String acceptLanguage, @Header("User-Agent") String userAgent);
```
daschult: @HeaderParam, right?

jianghaolu: Right.

PR #190, opened 2017-08-10
Run polling in immediate scheduler

jianghaolu:


azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java
```diff
@@ -154,7 +154,8 @@ public void setLongRunningOperationRetryTimeout(int longRunningOperationRetryTim
      * @return the observable of which a subscription will lead PUT or PATCH action.
      */
     public <T> Single<PollingState<T>> beginPutOrPatchAsync(Observable<Response<ResponseBody>> observable, final Type resourceType) {
-        return observable.map(new Func1<Response<ResponseBody>, PollingState<T>>() {
+        return observable
+                .observeOn(Schedulers.immediate()).map(new Func1<Response<ResponseBody>, PollingState<T>>() {
```
daschult: What does scheduling on the immediate scheduler do differently than the default?

jianghaolu: It's not necessary actually

PR #191, opened 2017-08-11
Adding content type annotations [ResponseContentType, RequestContentType]

anuchandy:
\\cc @jianghaolu @daschult - PR contains the ideas we discussed around content-type for encoding and decoding. Check the java-doc and share your thoughts.


client-runtime/src/main/java/com/microsoft/rest/v2/annotations/RequestContentType.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Defines the content type (MIME type) for the request body.
+ *
+ * If the Content-Type header present in {@link Headers} annotation then this annotation override it because this annotation is more specialized.
```
daschult: "This annotation will override a Content-Type string in a {@link Headers} annotation."

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/RequestContentType.java
```diff
@@ -0,0 +1,32 @@
...
+ * If the Content-Type header is passed as parameter via {@link HeaderParam} annotation then {@link HeaderParam} takes precedence over this annotation,
```
daschult: "A {@link HeaderParam} annotation on an interface parameter will override this annotation."

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/RequestContentType.java
```diff
@@ -0,0 +1,32 @@
...
+ * because that is what user selected during runtime.
+ * The value of this annotation overrides "Content-Type" header added in any interceptors, and "Content-Type" header added by the HTTP itself.
```
daschult: I don't think this sentence applies anymore since we shouldn't be setting content-type in interceptors anymore.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/RequestContentType.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * The precedence is @Header("Content-Type") [dynamic] > @RequestContentType(value) [static] > @Headers("Content-Type: ..") [static]
```
daschult: Do we still have a @Header annotation? I thought we just have @Headers?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/RequestContentType.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Once the "Content-Type" is derived, value will be used to choose the converter to encode the user provided POJO object to send it over wire.
```
daschult: ...derived, THE value will...

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/RequestContentType.java
```diff
@@ -0,0 +1,32 @@
...
+ * Finally derived "Content-Type" become the value of "Content-Type" header in the request.
+ */
+@Target({ElementType.METHOD})            // The context in which annotation is applicable i.e. this annotation (RequestContentType) can be applied only to methods
+@Retention(RetentionPolicy.RUNTIME)      // Record this annotation in the class file and make it available during runtime.
+public @interface RequestContentType {
+    String value();
```
daschult: Javadoc for value() please.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/ResponseContentType.java
```diff
@@ -0,0 +1,28 @@
...
+ * If the Content-Type header present in the response then that takes precedence over this annotation.
+ *
+ * The precedence is Response.Header("Content-Type") > @ResponseContentType(value)
+ *
+ * Once the "Content-Type" is derived, value will be used to choose the converter to decode response body bytes to POJO object
+ */
+@Target({ElementType.METHOD})            // The context in which annotation is applicable i.e. this annotation (ResponseContentType) can be applied only to methods
+@Retention(RetentionPolicy.RUNTIME)      // Record this annotation in the class file and make it available during runtime.
+public @interface ResponseContentType {
+    String value();
```
daschult: Javadoc here please.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/RequestContentType.java
anuchandy: it should be `@HeaderParam("Content-Type")` not `@Header("Content-Type")`. We have `@Headers` (plural) which is applicable only in method level and `@HeaderParam` is applicable in parameter level.

`@Headers` used to define values statically. `@HeaderParam` is used to set values during runtime.

daschult:
@anuchandy Have you made any changes for this PR lately? It'd be nice to get it through so that I can update the RestProxy to handle these annotations.

anuchandy:
Closing this, this annotation is not relevant anymore.

PR #192, opened 2017-08-11
Fix checkstyle

daschult:
This pull request is really about fixing checkstyle issues in our v2 branch. It looks like some other changes may have gotten merged into master and not v2, so those are the other changes that are showing up.


daschult:
@jianghaolu Please take a look at this pull request.

daschult:
@anuchandy Please take a look.

PR #193, opened 2017-08-14
Annotation parser refactor

daschult:
This pull request rearranges the classes used for the Annotation parser/RestProxy. It also adds several tests.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/EncodedParameter.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/EncodedParameter.java
@@ -1,0 +1,29 @@
+package com.microsoft.rest.v2;
+
+public class EncodedParameter {
+    private final String name;
+    private final String encodedValue;
+
+    public EncodedParameter(String name, String encodedValue) {
+        this.name = name;
+        this.encodedValue = encodedValue;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public String getEncodedValue() {
+        return encodedValue;
+    }
+
+    public boolean equals(Object rhs) {
+        return rhs instanceof EncodedParameter ? equals((EncodedParameter) rhs) : false;
+    }
+
+    public boolean equals(EncodedParameter rhs) {
+        return rhs != null &&
+                 name.equals(rhs.name) &&
+                 encodedValue.equals(rhs.encodedValue);
+    }
+}
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/EncodedParameter.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/EncodedParameter.java
@@ -1,29 +1,73 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
 package com.microsoft.rest.v2;
 
+/**
+ * An EncodedParameter is the result of encoding a query parameter or header name/value pair for a
+ * HTTP request. It contains the query parameter or header name, plus the query parameter's value or
+ * header's value.
+ */
 public class EncodedParameter {
     private final String name;
     private final String encodedValue;
 
+    /**
+     * Create a new EncodedParameter using the provided parameter name and encoded value.
+     * @param name The name of the new parameter.
+     * @param encodedValue The encoded value of the new parameter.
+     */
     public EncodedParameter(String name, String encodedValue) {
         this.name = name;
         this.encodedValue = encodedValue;
     }
 
+    /**
+     * Get this parameter's name.
+     * @return The name of this parameter.
+     */
     public String getName() {
         return name;
     }
 
+    /**
+     * Get the encoded value for this parameter.
+     * @return The encoded value for this parameter.
+     */
     public String getEncodedValue() {
         return encodedValue;
     }
 
+    /**
+     * Get whether or not this value equals the provided rhs value.
+     * @param rhs The value to compare against this value.
+     * @return Whether or not this value equals the provided rhs value.
+     */
+    @Override
     public boolean equals(Object rhs) {
         return rhs instanceof EncodedParameter ? equals((EncodedParameter) rhs) : false;
     }
 
+    /**
+     * Get whether or not this value equals the provided rhs value.
+     * @param rhs The value to compare against this value.
+     * @return Whether or not this value equals the provided rhs value.
+     */
     public boolean equals(EncodedParameter rhs) {
-        return rhs != null &&
-                 name.equals(rhs.name) &&
-                 encodedValue.equals(rhs.encodedValue);
+        return rhs != null
+                && name.equals(rhs.name)
+                && encodedValue.equals(rhs.encodedValue);
+    }
+
+    /**
+     * Get the unique hash code for this value.
+     * @return The unique hash code for this value.
+     */
+    @Override
+    public int hashCode() {
+        return name.hashCode() ^ encodedValue.hashCode();
     }
 }
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/EncodedParameter.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/EncodedParameter.java
@@ -11,7 +11,7 @@ package com.microsoft.rest.v2;
  * HTTP request. It contains the query parameter or header name, plus the query parameter's value or
  * header's value.
  */
-public class EncodedParameter {
+class EncodedParameter {
     private final String name;
     private final String encodedValue;
 
@@ -20,7 +20,7 @@ public class EncodedParameter {
      * @param name The name of the new parameter.
      * @param encodedValue The encoded value of the new parameter.
      */
-    public EncodedParameter(String name, String encodedValue) {
+    EncodedParameter(String name, String encodedValue) {
         this.name = name;
         this.encodedValue = encodedValue;
     }
@@ -29,7 +29,7 @@ public class EncodedParameter {
      * Get this parameter's name.
      * @return The name of this parameter.
      */
-    public String getName() {
+    public String name() {
         return name;
     }
 
@@ -37,7 +37,7 @@ public class EncodedParameter {
      * Get the encoded value for this parameter.
      * @return The encoded value for this parameter.
      */
-    public String getEncodedValue() {
+    public String encodedValue() {
         return encodedValue;
     }
 
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
@@ -40,17 +40,17 @@ import java.util.Map;
 public class RestProxy implements InvocationHandler {
     private final String host;
     private final RestClient restClient;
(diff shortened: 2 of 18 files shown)


daschult:
@jianghaolu @anuchandy Can one of you take a look? This is just another reworking step for the example that Jianghao wrote for me, plus adding tests.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpClient.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package com.microsoft.rest.v2.http;
+
+import java.io.IOException;
+
+/**
+ * A generic interface for sending HTTP requests and getting responses.
+ */
+public interface HttpClient {
```
daschult: @RikkiGibson We need an implementation of this interface that uses RxNetty. We'll have to change the interface to support RxJava, but this is the starting point.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpResponse.java
```diff
@@ -0,0 +1,48 @@
...
+package com.microsoft.rest.v2.http;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This class contains all of the details necessary for reacting to a HTTP response from a
+ * HttpResponse.
+ */
+public interface HttpResponse {
```
daschult: @RikkiGibson We'll need an implementation of this one for RxNetty too.

client-runtime/src/main/java/com/microsoft/rest/v2/http/OkHttpClient.java
```diff
@@ -0,0 +1,56 @@
...
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
+import java.io.IOException;
+
+/**
+ * A HttpClient that is implemented using OkHttp.
+ */
+public class OkHttpClient implements HttpClient {
```
daschult: This is only temporary and will be deleted when we move to RxNetty.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -19,209 +20,169 @@
...
+            host = hostAnnotation.value();
+        }
+
+        for (Method method : swaggerInterface.getDeclaredMethods()) {
+            final SwaggerMethodProxyDetails methodProxyDetails = interfaceProxyDetails.getMethodProxyDetails(method.getName());
 
             if (method.isAnnotationPresent(GET.class)) {
-                info.method = "GET";
-                info.relativePath = method.getAnnotation(GET.class).value();
+                methodProxyDetails.setMethodAndRelativePath("GET", method.getAnnotation(GET.class).value());
```
anuchandy: There may be some reasons for doing the parsing of all annotation here, but shouldn't the parsing responsibility reside inside `SwaggerInterfaceProxyDetails`, i.e `SwaggerInterfaceProxyDetails` ctr can take `SwaggerInterface` reference and have parsing logic there. That way when i retrieve `MethodProxyDetails ` of a method, it is already populated.

anuchandy: If we decide to move the parsing logic to `SwaggerInterfaceProxyDetails`, should we do that on demand? i.e. when user invokes a method then only create `MethodProxyDetails ` and cache it.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -19,209 +20,169 @@
...
+                    else if (annotationType.equals(PathParam.class)) {
+                        final PathParam pathParamAnnotation = (PathParam) annotation;
+                        methodProxyDetails.addPathSubstitution(pathParamAnnotation.value(), parameterIndex, !pathParamAnnotation.encoded());
+                    }
+                    else if (annotationType.equals(QueryParam.class)) {
+                        final QueryParam queryParamAnnotation = (QueryParam) annotation;
+                        methodProxyDetails.addQuerySubstitution(queryParamAnnotation.value(), parameterIndex, !queryParamAnnotation.encoded());
+                    }
+                    else if (annotationType.equals(HeaderParam.class)) {
+                        final HeaderParam headerParamAnnotation = (HeaderParam) annotation;
```
anuchandy: We also have `headers` annotation which is a method level annotation. If the same header key present in `headers` and `headerParam` we need to decide which one take precedence.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -19,209 +20,169 @@
...
             return null;
-        } else if (method.getReturnType().isAssignableFrom(InputStream.class)) {
-            return response.body().byteStream();
-        } else if (method.getReturnType().isAssignableFrom(byte[].class)) {
-            return response.body().bytes();
+        } else if (returnType.isAssignableFrom(InputStream.class)) {
+            return response.getBodyAsInputStream();
+        } else if (returnType.isAssignableFrom(byte[].class)) {
+            return response.getBodyAsByteArray();
         } else {
```
jianghaolu: Check Observable type for async

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: I usually like to keep all of my parsing logic to a single file so that it's easier to keep it all in my head. However, I like the idea of lazily parsing a swagger interfaces since I'm sure that these interfaces can get pretty large and that may produce a performance impact. Let me make this change in a future pull request.

daschult: I've created this branch for the lazy parsing work: [LazilyParseSwaggerInterface](https://github.com/daschult/autorest-clientruntime-for-java/tree/LazilyParseSwaggerInterface)

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: Yeah, I remembered that after I put out this pull request. I implemented the Headers annotation during my async work (the next pull request I'll send out), and rather than having one take precedence over the other, I made is so that they are additive with each others. It's allowed to have multiple values assigned to a single header, so if we have both HeaderParam and Headers annotations, then I just add them together (separate them by a ",").

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpResponse.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    boolean hasBody();
+
+    /**
+     * Get this response object's body as an InputStream. If this response object doesn't have a
+     * body, then null will be returned.
+     * @return This response object's body as an InputStream. If this response object doesn't have a
+     * body, then null will be returned.
+     */
+    InputStream getBodyAsInputStream();
```
jianghaolu: `inputStream()`, `byteArray()`, `string()`?

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
jianghaolu: ok to be in a next PR

client-runtime/src/main/java/com/microsoft/rest/v2/http/UrlBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+    private String host;
+    private String path;
+    private String query;
+
+    /**
+     * Set the setScheme/protocol that will be used to build the final URL.
+     * @param scheme The setScheme/protocol that will be used to build the final URL.
+     * @return This UrlBuilder so that multiple setters can be chained together.
+     */
+    public UrlBuilder setScheme(String scheme) {
```
daschult: Updated method names in UrlBuilder.

client-runtime/src/main/java/com/microsoft/rest/v2/EncodedParameter.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package com.microsoft.rest.v2;
+
+/**
+ * An EncodedParameter is the result of encoding a query parameter or header name/value pair for a
+ * HTTP request. It contains the query parameter or header name, plus the query parameter's value or
+ * header's value.
+ */
+public class EncodedParameter {
```
daschult: That means just removing the public access modifier, right? If so, done.

jianghaolu: yes

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodProxyDetails.java
```diff
@@ -0,0 +1,233 @@
...
+    private final List<Substitution> pathSubstitutions = new ArrayList<>();
+    private final List<Substitution> querySubstitutions = new ArrayList<>();
+    private final List<Substitution> headerSubstitutions = new ArrayList<>();
+    private Integer bodyContentMethodParameterIndex;
+
+    /**
+     * Get the HTTP method that will be used to complete the Swagger method's request.
+     * @return The HTTP method that will be used to complete the Swagger method's request.
+     */
+    public String getMethod() {
```
daschult: Done.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: Yup.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpResponse.java
daschult: I want to make sure its clear that it's the body that is being returned as an InputStream, byte[], or String, so can we settle on just getting rid of "get" and leaving it as "bodyAsInputStream()"?

daschult:
@anuchandy @jianghaolu I just pushed out some changes based on your reviews. If you're happy with the changes, can you sign off on this request? I'll merge it after that.

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodProxyDetails.java
```diff
@@ -0,0 +1,233 @@
...
+     */
+    public String method() {
+        return method;
+    }
+
+    /**
+     * Set the HTTP method that will be used to complete the Swagger method's request.
+     * @param method The HTTP method that will be used to complete the Swagger method's request.
+     */
+    public void setMethod(String method) {
```
jianghaolu: Do you think it will be helpful to make these chaining withers?

daschult: Based on Anu's suggestion to parse the interface lazily, I don't think so. I'm going to be changing this class pretty drastically soon, so I wouldn't think about it too much.

jianghaolu: Sounds good

PR #196, opened 2017-08-16
More definitive resource id matching

jianghaolu:


azure-client-runtime/src/main/java/com/microsoft/azure/credentials/AzureTokenCredentials.java
```diff
@@ -39,15 +41,18 @@ public AzureTokenCredentials(AzureEnvironment environment, String domain) {
...
     @Override
     protected final String getToken(Request request) throws IOException {
-        String host = request.url().host();
-        for (String endpoint : environment().endpoints().values()) {
-            if (host.contains(endpoint)) {
-                // Remove leading dots
-                host = endpoint.replaceAll("^\\.*", "");
-                break;
+        String host = request.url().toString().toLowerCase();
+        String resource = environment().activeDirectoryResourceId();
```
daschult: Does this resource variable only hold the scheme and host of the URL, or does it have more (a path, query, or fragment)? If that's the case, should you try to keep those additional pieces on lines 49 and 52?

jianghaolu: Just the scheme and host.

jianghaolu:
Merging to unblock release.

PR #197, opened 2017-08-16
Make http client async

daschult:
Allow RestProxy to work with asynchronous swagger interface methods. I made a big assumption that our swagger interfaces will always have a synchronous and asynchronous version of a particular REST API. By making this assumption, I can determine what the parameterized type of the asynchronous return type will be. For example, if we have a method with the signature:
```java
Single<VirtualMachine> getVirtualMachineAsync()
```
Then I assume that I will find a method with this signature too:
```java
VirtualMachine getVirtualMachine()
```


daschult:
I made this pull request against master instead of v2. Oops.

PR #198, opened 2017-08-16
Make http client async

daschult:
Allow RestProxy to work with asynchronous swagger interface methods. I made a big assumption that our swagger interfaces will always have a synchronous and asynchronous version of a particular REST API. By making this assumption, I can determine what the parameterized type of the asynchronous return type will be. For example, if we have a method with the signature:
```java
Single<VirtualMachine> getVirtualMachineAsync()
```
Then I assume that I will find a method with this signature too:
```java
VirtualMachine getVirtualMachine()
```


RikkiGibson:
Travis CI says:
```
[ERROR] /home/travis/build/Azure/autorest-clientruntime-for-java/client-runtime/src/test/java/com/microsoft/rest/v2/http/UrlBuilderTests.java:[7,8] class URLBuilderTests is public, should be declared in a file named URLBuilderTests.java
```
Probably unrelated to your work.

daschult:
I'll look at that failure in a couple of minutes.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpHeaders.java
```diff
@@ -0,0 +1,40 @@
...
+package com.microsoft.rest.v2.http;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * A collection of headers that will be applied to a HTTP request.
+ */
+public class HttpHeaders implements Iterable<HttpHeader> {
```
RikkiGibson: RxNetty really wants to consume `Map<? extends CharSequence, Iterable<? extends Object>>`-- it copies the request object each time you add a header individually. May not necessitate changing things around but worth noting.

client-runtime/src/main/java/com/microsoft/rest/v2/http/UrlBuilder.java
```diff
@@ -19,8 +19,9 @@
     private String query;
 
     /**
-     * Set the withScheme/protocol that will be used to build the final URL.
-     * @param scheme The withScheme/protocol that will be used to build the final URL.
+<<<<<<< HEAD
```
RikkiGibson: Is this a merge conflict marker?

daschult: Yup. Removed.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpHeaders.java
daschult: I'd recommend doing a conversion within your implementation of HttpClient.sendRequestAsync(). That seems like a specific thing about RxNetty and not a generic concept that all HttpClient implementations would need. Thanks for the heads up about it though.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -180,6 +250,24 @@ else if (annotationType.equals(BodyParam.class)) {
                     }
                 }
             }
+
+            final Class<?> returnType = method.getReturnType();
+            final boolean isAsync = (returnType == Single.class || returnType == Completable.class);
```
jianghaolu: And `Observable`

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -180,6 +250,24 @@ else if (annotationType.equals(BodyParam.class)) {
...
+            methodProxyDetails.setIsAsync(isAsync);
+            if (!isAsync) {
+                methodProxyDetails.setReturnType(returnType);
+            }
+            else {
+                final String asyncMethodName = method.getName();
```
jianghaolu: Trying to find its corresponding sync method? Why is this needed?

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpResponse.java
```diff
@@ -6,43 +6,75 @@
...
 
     /**
      * Get this response object's body as a byte[]. If this response object doesn't have a body,
      * then null will be returned.
      * @return This response object's body as a byte[]. If this response object doesn't have a body,
      * then null will be returned.
-     * @throws IOException On network issues.
      */
-    byte[] bodyAsByteArray() throws IOException;
+    public abstract Single<byte[]> bodyAsByteArrayAsync();
```
jianghaolu: @anuchandy @RikkiGibson Could Netty return multiple bytebufs for one response?

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -84,19 +94,61 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
+            final Single<? extends HttpResponse> asyncResponse = httpClient.sendRequestAsync(request);
+            final Class<?> methodReturnType = method.getReturnType();
+            if (methodReturnType.equals(Single.class)) {
+                result = asyncResponse.flatMap(new Func1<HttpResponse, Single<?>>() {
+                    @Override
+                    public Single<?> call(HttpResponse response) {
+                        Single<?> asyncResult;
+                        final Class<?> singleReturnType = methodDetails.getReturnType();
+                        if (methodDetails.method().equalsIgnoreCase("HEAD")) {
+                            asyncResult = Single.just(null);
```
jianghaolu: In Azure, if the response is boolean, we return true for 2xx responses and false otherwise. You can also do this in the Azure flavored parser.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpResponse.java
anuchandy: yes, if the response is chunked netty can return multiple `HttpContent`s, in that case this should be an `Observable<byte[]>`. If we insert the [HttpObjectAggregator](https://netty.io/4.0/api/io/netty/handler/codec/http/HttpObjectAggregator.html) handler in the pipeline then these chunks will be merged into a single `HttpContent`.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: This is needed because we need to know the return type for the asynchronous method's parameterized type (T in Single<T> or Observable<T>) for deserialization.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpResponse.java
RikkiGibson: Yeah, for file download type use cases we may want to expose `Observable<ByteBuf>` or `Observable<byte[]>` depending on the copying semantics.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: Done.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: Okay. Let's talk more about that when I start working on the Azure flavored parser.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpResponse.java
daschult: It seems to me that if was consuming this interface, then I would expect bodyAsByteArrayAsync() to return the entire response's body as a single byte array. If we want to provided the "user" (the user of these interfaces is only our proxy implementations of our "inner" interfaces, right?) with the option of getting the byte array as chunks, such as bodyAsSegmentedByteArrayAsync() or bodyAsChunkedByteArrayAsync().

daschult:
Let me know if there was anything that you still wanted me to change and I'll fix it in a follow up pull request.

PR #199, opened 2017-08-18
Add mock http client

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodProxyDetails.java
```diff
@@ -233,7 +233,7 @@ public void setReturnType(Class<?> returnType) {
      * object that is returned from the method.
      * @return The synchronous return type for the method that this object describes.
      */
-    public Class<?> getReturnType() {
+    public Class<?> returnType() {
```
jianghaolu: We may need to use Type instead of Class<?> since Class<?> has difficulty handling generic types. See https://github.com/Azure/autorest-clientruntime-for-java/blob/master/client-runtime/src/main/java/com/microsoft/rest/serializer/JacksonAdapter.java#L127.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpHeaders.java
```diff
@@ -33,6 +49,18 @@ public HttpHeaders add(String headerName, String headerValue) {
...
+    /**
+     * Get the header value for the provided header name. If the header name isn't found, then null
+     * will be returned.
+     * @param headerName The name of the header to look for.
+     * @return The String value of the header, or null if the header isn't found.
+     */
+    public String get(String headerName) {
+        final String headerKey = headerName.toLowerCase();
+        final HttpHeader header = headers.get(headerKey);
+        return header == null ? null : header.getValue();
```
anuchandy: Since a header can have multiple values should we return `List<String>` or user needs split it by comma? If we return List<String> then it should be read-only as the only way to add header value is via `HttpHeaders.add(.., ..)`.

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodProxyDetails.java
daschult: Type is an interface that Class implements, so it doesn't necessarily give us any more information to use Type instead of Class (as far as I understand). Also, I used a debugger to step through our unit tests and I wasn't able to get more generic type information by casting it to other Type classes (ParameterizedType, Type, etc.). Maybe we should talk some more about this though if you know more about it.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpHeaders.java
daschult: I find it really frustrating when I know that a collection has a single value in it, but there's no easy way to get the single value without iterating through the collection. For this reason, I've left the original get() method (renamed to getValue()) and I've added getValues() that returns a String[].

PR #200, opened 2017-08-21
Lazily parse swagger interface

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -46,58 +32,45 @@
...
+            final String bodyContentString = serializer.serialize(bodyContentObject);
+            request.withBody(bodyContentString, mimeType);
         }
 
-        Object result = null;
-        final Type returnType = methodDetails.returnType();
+        Object result;
+        final Type returnType = methodParser.returnType();
         final TypeToken returnTypeToken = TypeToken.of(returnType);
         if (returnTypeToken.isSubtypeOf(Completable.class)) {
```
RikkiGibson: What's the plan for handling `Observable<>` return types in service interfaces?

RikkiGibson:
@anuchandy Any thoughts on this PR?

client-runtime/pom.xml
```diff
@@ -68,6 +68,31 @@
       <artifactId>okhttp-urlconnection</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.reactivex</groupId>
+      <artifactId>rxnetty-http</artifactId>
+      <version>0.5.2</version>
```
jianghaolu: Don't specify versions in child modules.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -46,58 +32,45 @@
...
 
     @Override
     public Object invoke(Object proxy, final Method method, Object[] args) throws Throwable {
-        final String methodName = method.getName();
-        final SwaggerMethodProxyDetails methodDetails = interfaceDetails.getMethodProxyDetails(methodName);
-
-        final String actualHost = methodDetails.applyHostSubstitutions(host, args);
-        final String[] hostParts = actualHost.split("://");
-        final String actualPath = methodDetails.getSubstitutedPath(args);
+        final SwaggerMethodParser methodParser = interfaceParser.methodParser(method);
```
jianghaolu: No we don't want lazy parsing - calling the method should be fast. We should understand everything about the interface when the proxy is built and cache as much as possible.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: I spent some time trying to think about how to properly parse an Observable<> return type, but I just couldn't think of the right way to do it. I think that was because I'm not sure there's an appropriate use case. Paging may be a good example, but I need to see an example of a paging request before I code that up.
Since that will be an easy thing to add later, I'm fine with postponing it until then.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: I understand your concern about performance, but this should actually be faster. If we do everything all at once, then we're probably going to be doing a lot of extra work that the user won't need. I imagine they'll only call a few methods on a given interface, in which case it would be better to only parse the ones they request. Once we parse a method, we do cache the parse so the next time it will just be a lookup.

client-runtime/pom.xml
daschult: Moved versions to root pom.xml.

PR #203, opened 2017-08-23
[Do not merge] RestProxy support in azure-client-runtime

RikkiGibson:
I'm making some changes in azure-client-runtime to support RestProxy in azure-sdk-for-java. These are fairly hacky early-stage changes to try and keep all the OkHttp-based stuff compiling while providing us a path forward to full RestProxy support eventually.

Please let me know what you think about the changes.


client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyClient.java
```diff
@@ -64,10 +73,21 @@ private SSLEngine getSSLEngine(String host) {
...
 
+        // Is caching of rxnetty's HttpClient preferred for performance?
         boolean isSecure = "https".equalsIgnoreCase(uri.getScheme());
         io.reactivex.netty.protocol.http.client.HttpClient<ByteBuf, ByteBuf> rxnClient =
                 io.reactivex.netty.protocol.http.client.HttpClient.newClient(uri.getHost(), isSecure ? 443 : 80);
 
+        for (int i = 0; i < handlerConfigs.length; i++) {
+            ChannelHandlerConfig config = handlerConfigs[i];
+            if (config.mayBlock()) {
+                EventExecutorGroup executorGroup = new DefaultEventExecutorGroup(1);
```
anuchandy: minor: Instead of creating one EventLoopGroup per handlers those belongs to `may-block` set, we could create one `DefaultEventExecutorGroup` with a fixed number of EventLoops (e.g. 10) and share it, that way threads will be pooled for `may-block` handlers.

azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceClient.java
```diff
@@ -49,6 +56,20 @@ protected AzureServiceClient(String baseUrl, ServiceClientCredentials credential
      */
     protected AzureServiceClient(RestClient restClient) {
         super(restClient);
+        rpHttpClient = new RxNettyClient(new ChannelHandlerConfig(new Func0<ChannelHandler>() {
+            @Override
+            public ChannelHandler call() {
+                return new AzureTokenCredentialsHandler((AzureTokenCredentials) AzureServiceClient.this.restClient().credentials());
```
daschult: Is there someway to get to credentials() without using restClient()?

azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceClient.java
```diff
@@ -49,6 +56,20 @@ protected AzureServiceClient(String baseUrl, ServiceClientCredentials credential
...
+            }
+        }, true));
+    }
+
+    private final HttpClient rpHttpClient;
```
daschult: I know that "rp" stands for RestProxy (it does, right?), but no one else will. Just call it "httpClient" for now.

azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceClient.java
```diff
@@ -49,6 +56,20 @@ protected AzureServiceClient(String baseUrl, ServiceClientCredentials credential
...
+    /**
+     * @return the RestProxy HTTP client.
+     */
+    public HttpClient rpHttpClient() {
```
daschult: Same as above. Change to "httpClient".

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureTokenCredentialsHandler.java
```diff
@@ -0,0 +1,26 @@
+package com.microsoft.azure.v2;
+
+import com.microsoft.azure.credentials.AzureTokenCredentials;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelOutboundHandlerAdapter;
+import io.netty.channel.ChannelPromise;
+import io.netty.handler.codec.http.HttpRequest;
+
+public class AzureTokenCredentialsHandler extends ChannelOutboundHandlerAdapter {
```
daschult: This appears to be a Netty specific handler. If I have classes like that, I usually add the technology that they are specific to either in their package or in their class name, especially since ChannelOutboundHandlerAdapter isn't clear that its a Netty class.

azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceClient.java
daschult: Also, will we create a new AzureTokenCredentialsHandler for each outgoing HTTP request? If so, that should be pulled out into a member variable of AzureServiceClient.

client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyClient.java
```diff
@@ -64,10 +73,21 @@ private SSLEngine getSSLEngine(String host) {
...
+                rxnClient = rxnClient.addChannelHandlerLast(executorGroup, "az-client-handler-" + i, config.factory());
+            } else {
+                rxnClient = rxnClient.addChannelHandlerLast("az-client-handler-" + i, config.factory());
```
daschult: Calculate "az-client-handler-" + I and config.factory() before the if statement. That way they'll be easier to change for both branches of the if statement.

client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyClient.java
RikkiGibson: If desired, down the line we will be able to keep an EventExecutorGroup on the RxNettyClientAdapter instance with minimal effects outside the class.

azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceClient.java
RikkiGibson: Let's figure something out for credentials. Probably one of the more hacked up parts of this.

RikkiGibson: Yeah, the handler is getting re-instantiated but doesn't really have any of its own state beyond the credentials which are already shared. It could be a final local or an instance variable, I think.

azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceClient.java
RikkiGibson: Naming conflict with a base class member-- perhaps I'll rename the original `httpClient` to `okHttpClient` and this to merely `httpClient`?

RikkiGibson:
Keeping compatibility with all the OkHttp stuff while we try to transition over to RestProxy is just tricky. I don't see much of an alternative to keeping all the OkHttp/Retrofit stuff around while we get more things moved over. We gotta be able to build this stuff.

PR #204, opened 2017-08-25
Add Adapter suffix to RxNettyClient and OkHttpClient

RikkiGibson:


client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -13,7 +13,7 @@
 import com.microsoft.rest.v2.http.HttpHeader;
 import com.microsoft.rest.v2.http.HttpRequest;
 import com.microsoft.rest.v2.http.HttpResponse;
-import com.microsoft.rest.v2.http.OkHttpClient;
+import com.microsoft.rest.v2.http.OkHttpClientAdapter;
```
daschult: Rather than having the classes be <implementation>ClientAdapter, just have them be <implementation>Adapter.

RikkiGibson: Good call.

RikkiGibson:
Rename is done, will probably merge after CI @daschult

RikkiGibson:
The commit doing the rename has an inaccurate message. The types are actually named OkHttpAdapter and RxNettyAdapter.

PR #205, opened 2017-08-25
Using Heap based buffer explicitly for response contents

anuchandy:
`Unpooled.buffer(contentLength)` may return a direct buffer, let's use the pooled heap buffer explicitly for collecting the chunks.


daschult:
Why are the checks on this branch failing? Are you still working on this? Once the checks pass, I'll take a look.

anuchandy:
@daschult thanks, fixed the style errors.

Issue #210, opened 2017-09-06
[RestProxy] Non-string request bodies


RikkiGibson:
File upload is the obvious case for this. Perhaps request bodies should be ByteBuf with a MIME type denoted somewhere. I believe we expect the largest size for a single request body to be on the order of 10 MB.

PR #211, opened 2017-09-08
Make HttpRequest take InputStream instead of String body

daschult:


daschult:
Fixes https://github.com/Azure/autorest-clientruntime-for-java/issues/210.

client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyAdapter.java
```diff
@@ -42,52 +47,91 @@ private SSLEngine getSSLEngine(String host) {
...
-            rxnClient = rxnClient.secure(getSSLEngine(uri.getHost()));
+            result = obsResponse
+                    .map(new Func1<HttpClientResponse<ByteBuf>, HttpResponse>() {
+                        @Override
+                        public HttpResponse call(HttpClientResponse<ByteBuf> rxnRes) {
+                            return new RxNettyResponse(rxnRes);
+                        }
+                    })
+                    .toSingle();
+        } catch (URISyntaxException e) {
```
RikkiGibson: Why expand the try block to include a bunch of stuff that won't throw that exception? Do you prefer to avoid early returns?

daschult: I do generally not like early returns. I find it easier to understand code if it has one entry and one exit. In this case, I really wanted to combine the Single.error() call so that any exception would go to the same place and would skip the rest of the logic inside the try...catch block.

client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyAdapter.java
```diff
@@ -89,7 +91,7 @@ public HttpResponse call(HttpClientResponse<ByteBuf> rxnRes) {
                         }
                     })
                     .toSingle();
-        } catch (URISyntaxException e) {
+        } catch (URISyntaxException | IOException e) {
```
RikkiGibson: Neat!

client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyAdapter.java
```diff
@@ -42,52 +47,93 @@ private SSLEngine getSSLEngine(String host) {
...
+        protected InputStream generateState() {
+            return this.is;
+        }
+
+        @Override
+        protected InputStream next(InputStream state, Observer<? super byte[]> observer) {
+            byte[] buffer = new byte[size];
+            try {
+                int count = state.read(buffer);
+                if (count == -1)
```
RikkiGibson: CheckStyle insists these should have braces.

client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyAdapter.java
```diff
@@ -42,52 +47,93 @@ private SSLEngine getSSLEngine(String host) {
...
+                    observer.onCompleted();
+                else if (count < size)
+                    observer.onNext(Arrays.copyOf(buffer, count));
```
anuchandy: [Adding this for future reference]: In case we find that we want to optimize by avoiding making additional copy `Arrays.copyOf` then we could use `java.nio.ByteBuffer` instead of `byte[..]` 

```java
byte[] buffer = new byte[size]; ==> java.nio.ByteBuffer buffer = ByteBuffer.allocate(size);
int count = state.read(buffer);  ==> int count = state.read(buffer.array());
Arrays.copyOf(buffer, count)   ==> ByteBuffer.wrap(buffer.array(), 0, count).array();
```

client-runtime/src/main/java/com/microsoft/rest/v2/http/RxNettyAdapter.java
```diff
@@ -42,52 +47,93 @@ private SSLEngine getSSLEngine(String host) {
...
+            return this.is;
+        }
+
+        @Override
+        protected InputStream next(InputStream state, Observer<? super byte[]> observer) {
+            byte[] buffer = new byte[size];
+            try {
+                int count = state.read(buffer);
+                if (count == -1)
+                    observer.onCompleted();
```
anuchandy: Should we close the inputstream on completion & on error?

RikkiGibson: Not sure--is this method lifted from rxjava-string?

daschult:
Closing and opening to try to refresh travis.

RikkiGibson:
Checkstyle complains about missing header comments and Javadocs in HttpRequestBody.. did local Checkstyle miss that?

I'm tempted to just merge to v2 in order to move things forward.

daschult:
Yeah, that seems like a good idea. I’ll fix all of my checkstyle issues and then let’s just merge the branch.

I’ll just merge it when I’m done.

Issue #206, opened 2017-08-25
RestProxy and encoded query parameters


RikkiGibson:
RestProxy handles the `encoded = true` argument to PathParam and QueryParam by calling URLEncoder.encode. We may have problems getting the right behavior with both path parameters and query parameters because they're encoded differently and URLEncoder can't tell whether the thing you're encoding is a query argument or a path component.

RikkiGibson:
Guave to the rescue again--we already have this class in the project: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/net/UrlEscapers.html

RikkiGibson:
Should probably write some tests about exercising URL encoding.

daschult:
Rikki already has a branch for doing this work: [https://github.com/Azure/autorest-clientruntime-for-java/tree/URLEscapers](https://github.com/Azure/autorest-clientruntime-for-java/tree/URLEscapers)

RikkiGibson:
@daschult I'm actually hung up on this test--would appreciate you taking a look if you have the time. It looks as though the URL path and query parameters are being double encoded. Every place I'd expect to see %, I see %25, which is the encoding for the literal '%' character. I have no idea how the double encoding could be happening-- the debugger seems to encode the path segment once, and the query parameter once.

daschult:
I resolved this by not using java.net.URI to compose our urls anymore. We're doing it by hand. If this becomes a problem, then we'll need to revisit this issue.

PR #212, opened 2017-09-08
Change UrlBuilder to not encode on toString()

daschult:


daschult:
Resolves https://github.com/Azure/autorest-clientruntime-for-java/issues/206

Issue #202, opened 2017-08-23
RestProxy and expected response codes


RikkiGibson:
Swagger specifies what response codes are valid for a method and what they mean. 

e.g. https://github.com/Azure/azure-rest-api-specs/blob/current/specification/compute/resource-manager/Microsoft.Compute/2017-03-30/compute.json#L494

```
        "responses": {
          "200": {
            "description": "OK",
            "schema": {
              "$ref": "#/definitions/OperationStatusResponse"
            }
          },
          "202": {
            "description": "Accepted"
          },
          "204": {
            "description": "No Content"
          }
        }
```

The different response codes can have different return types associated with them. I'm guessing that there's usually one return type or nothing--some kind of in-progress value is given back through a header or something. Right now, we have generated code in azure-sdk-for-java for each service method that validates those response codes and otherwise wraps up the response body in an exception.

Perhaps there should be an annotation on a service method which indicates what response codes are acceptable? It would look something like this:

```
interface UsagesService {
    @Responses({ 200, 202, 204 })
    @Headers({ "Content-Type: application/json; charset=utf-8" })
    @GET("subscriptions/{subscriptionId}/providers/Microsoft.Storage/usages")
    Observable<PageImpl<UsageInner>> list(
        @PathParam("subscriptionId") String subscriptionId,
        @QueryParam("api-version") String apiVersion,
        @HeaderParam("accept-language") String acceptLanguage,
        @HeaderParam("User-Agent") String userAgent);
}
```

Does this handle our use case?

RikkiGibson:
Leaning toward thinking it's probably not necessary to have annotations for this, as the expected response codes across Azure management are pretty consistent (either 200, or 200/202/204 where only 200 returns an entity)

anuchandy:
We need to register the models for errors as there can be multiple model for different error codes and these error models may not be same across different swaggers.

RikkiGibson:
I suggest having a default behavior of CloudException for any non-2xx response code, and then something like an annotation on each method for each type of custom exception that can be thrown. @daschult this is probably the main thing that RestProxy can start doing which will make `ResponseBuilder` and `ServiceResponse` in the runtime which depend on Retrofit stuff unnecessary.

```
interface MyService {
    @ErrorResponse(405, CustomErrorType.class)
    @ErrorResponse(400, OtherErrorType.class)
    Entity myMethod(...)
}

RikkiGibson:
[...]

PR #213, opened 2017-09-11
Add ExpectedResponse and UnexpectedResponseExceptionType annotations

daschult:
This pull request addresses [this issue](https://github.com/Azure/autorest-clientruntime-for-java/issues/202).

I added ExpectedResponse and UnexpectedResponseExceptionType annotations. I chose to use UnexpectedResponseExceptionType instead of ErrorResponse because I felt that it better described what the annotation was actually doing.

I also added some tests to improve test coverage and to ensure that RestProxy was doing what we expected.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/RestException.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/RestException.java
@@ -6,6 +6,7 @@
 
 package com.microsoft.rest;
 
+import com.microsoft.rest.v2.http.HttpResponse;
 import okhttp3.ResponseBody;
 import retrofit2.Response;
 
@@ -16,7 +17,12 @@ public class RestException extends RuntimeException {
     /**
      * Information about the associated HTTP response.
      */
-    private Response<ResponseBody> response;
+    private Response<ResponseBody> responseOkHttp;
+
+    /**
+     * Information about the associated HTTP response.
+     */
+    private HttpResponse responseV2;
 
     /**
      * The HTTP response body.
@@ -31,7 +37,7 @@ public class RestException extends RuntimeException {
      */
     public RestException(String message, Response<ResponseBody> response) {
         super(message);
-        this.response = response;
+        this.responseOkHttp = response;
     }
 
     /**
@@ -43,7 +49,31 @@ public class RestException extends RuntimeException {
      */
     public RestException(String message, Response<ResponseBody> response, Object body) {
         super(message);
-        this.response = response;
+        this.responseOkHttp = response;
+        this.body = body;
+    }
+
+    /**
+     * Initializes a new instance of the RestException class.
+     *
+     * @param message the exception message or the response content if a message is not available
+     * @param response the HTTP response
+     */
+    public RestException(String message, HttpResponse response) {
+        super(message);
+        this.responseV2 = response;
+    }
+
+    /**
+     * Initializes a new instance of the RestException class.
+     *
+     * @param message the exception message or the response content if a message is not available
+     * @param response the HTTP response
+     * @param body the deserialized response body
+     */
+    public RestException(String message, HttpResponse response, Object body) {
+        super(message);
+        this.responseV2 = response;
         this.body = body;
     }
 
@@ -51,7 +81,14 @@ public class RestException extends RuntimeException {
      * @return information about the associated HTTP response
      */
     public Response<ResponseBody> response() {
-        return response;
+        return responseOkHttp;
+    }
+
+    /**
+     * @return information about the associated HTTP response
+     */
+    public HttpResponse responseV2() {
+        return responseV2;
     }
 
     /**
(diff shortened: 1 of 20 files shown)


client-runtime/src/main/java/com/microsoft/rest/RestException.java
```diff
@@ -16,7 +17,12 @@
     /**
      * Information about the associated HTTP response.
      */
-    private Response<ResponseBody> response;
+    private Response<ResponseBody> responseOkHttp;
```
RikkiGibson: Smells like something that's too entrenched to factor out right away?

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/ExpectedResponse.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Example:
+ *   {@literal @}ExpectedResponse({200, 201})
+ *   {@literal @}POST("subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CustomerInsights/hubs/{hubName}/images/getEntityTypeImageUploadUrl")
+ *   void getUploadUrlForEntityType(@Path("resourceGroupName") String resourceGroupName, @Path("hubName") String hubName, @Path("subscriptionId") String subscriptionId, @Body GetImageUploadUrlInputInner parameters);
+ *
+ */
+@Retention(RUNTIME)
+@Target(METHOD)
+public @interface ExpectedResponse {
```
RikkiGibson: Consider pluralizing the name

client-runtime/src/main/java/com/microsoft/rest/RestException.java
daschult: Correct. I tried to remove this, but then I started getting errors in azure-client-runtime. I don't want to start messing with that package quite yet, so I'm leaving this here for now.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -109,6 +112,28 @@ else if (returnTypeToken.isSubtypeOf(Single.class)) {
...
+                    final Class<?> exceptionBodyType = methodParser.exceptionBodyType();
+                    final Constructor<? extends RestException> exceptionConstructor = exceptionType.getConstructor(String.class, HttpResponse.class, exceptionBodyType);
+
+                    String responseContent = null;
+                    try {
+                        responseContent = response.bodyAsString();
+                    } catch (IOException ignored) {
+                    }
+
+                    final Object exceptionBody = responseContent == null ? null : serializer.deserialize(responseContent, exceptionBodyType);
```
anuchandy: If there is an exception thrown by de-serializer while parsing the string `responseContent`, can we ensure message part of `IOException` contains `responseContent`?

anuchandy: i mean the message part of `IOException` we throw below (line#133)

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -109,6 +112,28 @@ else if (returnTypeToken.isSubtypeOf(Single.class)) {
...
+
+                    throw exceptionConstructor.newInstance("Status code " + responseStatusCode + ", " + responseContent, response, exceptionBody);
+                } catch (IllegalAccessException | InstantiationException | InvocationTargetException | NoSuchMethodException e) {
+                    throw new IOException("Status code " + responseStatusCode + ", but an instance of " + exceptionType.getCanonicalName() + " cannot be created.", e);
+                }
+            }
+
```
anuchandy: The http status code `204` should not contain a body, so either we may want to skip deserialization in 204 case or see the behavior of `serializer.deserialize(null /** responseBodyString** /, returnType)`, if it throws on null then it's a problem. what do you think?

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: I've added responseContent to the IOException's message if the responseContent isn't null or empty.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: On the line where I deserialize, I check if responseContent is null and only deserialize if it isn't null. I just added a check for if it is empty as well. I agree that we should be careful of NullPointerExceptions, but I don't think we'll have that problem here.

daschult:
@RikkiGibson @anuchandy Let me know if you guys have any more comments/concerns. If not, then I'll merge this once my CI passes.

anuchandy:
LGTM!

PR #214, opened 2017-09-11
Add RequestPolicy and begin migrating away from OkHttp/Retrofit

RikkiGibson:
This PR is pretty large and I apologize for that. The big beats of what this does includes:

- Removes usages of OkHttp and Retrofit across the most referenced types in the runtime, e.g. RestClient.
  * RestClient is really just a way to share configuration between different services and reuse an existing configuration as a base while making modifications. We can keep considering whether to keep it in its current form but it's looking pretty likely that a configuration object for service clients will be useful.
- Refactors interceptors in the runtime to use a RequestPolicy chaining concept which depends on the abstract HttpClient in the runtime.
  * Credentials are a big part of this-- all the Credentials objects no longer know what an Interceptor or a RequestPolicy or anything are and are just tasked with providing a value for the Authorization header. There don't appear to be tests in the runtime itself for AzureTokenCredentials and probably other classes, so it may take some work elsewhere to verify that they're actually working. This work might take place as part of refactoring https://github.com/azure/autorest.java to use the new runtime or it might be possible to integration test some generic part of Azure to verify that valid credentials are being attached to requests.


azure-client-authentication/src/main/java/com/microsoft/azure/credentials/RefreshTokenClient.java
```diff
@@ -32,7 +30,8 @@
     private final RefreshTokenService service;
 
     RefreshTokenClient(String baseUrl, Proxy proxy) {
-        OkHttpClient.Builder builder = new OkHttpClient.Builder().addInterceptor(new LoggingInterceptor(LogLevel.BODY_AND_HEADERS));
```
daschult: Are we not logging body and headers any more?

azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java
```diff
@@ -232,7 +233,6 @@ public void setLongRunningOperationRetryTimeout(int longRunningOperationRetryTim
         pollingState.withResourceType(resourceType);
         pollingState.withSerializerAdapter(restClient().serializerAdapter());
         return Observable.just(true)
-                .subscribeOn(Schedulers.io())
```
daschult: Why aren't we running this on the IO scheduler anymore?

azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceClient.java
```diff
@@ -49,6 +49,28 @@ protected AzureServiceClient(String baseUrl, ServiceClientCredentials credential
      */
     protected AzureServiceClient(RestClient restClient) {
         super(restClient);
+
+        final ServiceClientCredentials credentials = restClient().credentials();
+        final RxNettyAdapter rxnClient = new RxNettyAdapter();
+        httpClient = new RequestPolicyChain(new CredentialsPolicy.Factory(credentials), new RequestPolicy.Factory() {
```
daschult: I'm curious about the name RequestPolicyChain, but I'll keep going through this review to see if it makes more sense.

client-runtime/src/main/java/com/microsoft/rest/RestClient.java
```diff
@@ -6,241 +6,191 @@
...
+        this.serializerAdapter = builder.serializerAdapter;
+        this.responseBuilderFactory = builder.responseBuilderFactory;
+        this.credentials = builder.credentials;
+        this.logLevel = builder.logLevel;
+
+        this.customPolicyFactories = builder.customPolicyFactories;
+        this.fullPolicyChain = createPolicyChain();
+    }
+
+    private RequestPolicyChain createPolicyChain() {
```
daschult: Why is this called RequestPolicyChain? I understand the chain idea, but what are the "policies"?

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpClient.java
```diff
@@ -6,14 +6,16 @@
...
 import rx.Single;
+import rx.functions.Func1;
 
 import java.io.IOException;
 
 /**
  * A generic interface for sending HTTP requests and getting responses.
  */
-public abstract class HttpClient {
+public abstract class HttpClient implements RequestPolicy {
```
daschult: I'm a little wary about this. I'm not entirely certain what RequestPolicy is supposed to be, but I don't think that HttpClient should implement it. This feels like maybe the name "RequestPolicy" isn't right.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpHeaders.java
```diff
@@ -62,6 +62,19 @@ public HttpHeaders add(String headerName, String headerValue) {
     }
 
     /**
+     * Replace the value for the header named headerName,
+     * discarding any value previously added for that header.
+     * @param headerName The name of the header.
+     * @param headerValue The value of the header.
+     * @return This HttpHeaders instance.
+     */
+    public HttpHeaders overwrite(String headerName, String headerValue) {
```
daschult: I'd call it "set" instead of "overwrite". Overwrite is usually a flag that is passed to functions, not the name of a function.

client-runtime/src/main/java/com/microsoft/rest/v2/policy/RequestPolicyChain.java
```diff
@@ -0,0 +1,67 @@
...
+public class RequestPolicyChain extends HttpClient {
+    private final List<RequestPolicy.Factory> factories;
+
+    /**
+     * Creates RequestPolicyChain.
+     *
+     * @param factoryArray the factories that can creates RequestPolicy instances in the pipeline.
+     */
+    public RequestPolicyChain(RequestPolicy.Factory... factoryArray) {
+        factories = Arrays.asList(factoryArray);
```
daschult: Chain this constructor into the one on line 40 so that you only have to use Collections.reverse() once.

client-runtime/src/main/java/com/microsoft/rest/v2/policy/RequestPolicyChain.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Creates RequestPolicyChain.
+     *
+     * @param factories the factories that can creates RequestPolicy instances in the pipeline.
+     */
+    public RequestPolicyChain(List<RequestPolicy.Factory> factories) {
+        this.factories = new ArrayList<>(factories);
+        Collections.reverse(this.factories);
```
daschult: In fact, rather than reverse the collection here, why not reverse it in create()? That way you only call Collections.reverse() once, and it puts the reasoning for reversing the collection next to where the reversed collection will be used.

client-runtime/src/main/java/com/microsoft/rest/v2/policy/RetryPolicy.java
```diff
@@ -0,0 +1,77 @@
...
+    private RetryPolicy(int maxRetries, RequestPolicy next) {
+        this.maxRetries = maxRetries;
+        this.next = next;
+    }
+
+    private boolean shouldRetry(HttpResponse response) {
+        int code = response.statusCode();
+        //CHECKSTYLE IGNORE MagicNumber FOR NEXT 2 LINES
+        return tryCount < maxRetries
+                && (code == 408 || (code >= 500 && code != 501 && code != 505));
```
daschult: You might be able to make checkstyle happy and make this code easier to read if you extract these into static final variables with names. I have to admit that I don't remember what a 505 status code means.

client-runtime/src/main/java/com/microsoft/rest/v2/policy/SendRequestPolicyFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.microsoft.rest.v2.policy;
+
+import com.microsoft.rest.v2.http.HttpClient;
+
+/**
+ * A request policy factory which sends a request using an HTTP client.
+ * Generally should be placed at the end of a {@link RequestPolicyChain}.
+ */
+public class SendRequestPolicyFactory implements RequestPolicy.Factory {
```
daschult: What's the purpose of this class?

client-runtime/src/test/java/com/microsoft/rest/UserAgentTests.java
```diff
@@ -15,52 +22,57 @@
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 import okhttp3.Protocol;
-import okhttp3.Request;
 import okhttp3.Response;
-import retrofit2.Retrofit;
+import retrofit2.http.HTTP;
```
daschult: Didn't quite get rid of retrofit2?

azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java
RikkiGibson: I think I just merged master in and this work happened to come along.
https://github.com/Azure/autorest-clientruntime-for-java/pull/190

azure-client-authentication/src/main/java/com/microsoft/azure/credentials/RefreshTokenClient.java
RikkiGibson: This needs work. LoggingInterceptor is missing because it turned into LoggingPolicy.

client-runtime/src/main/java/com/microsoft/rest/v2/policy/RequestPolicyChain.java
RikkiGibson: .create() is expected to be called per-request.

client-runtime/src/test/java/com/microsoft/rest/UserAgentTests.java
RikkiGibson: Whoops, will take another look at that one.

RikkiGibson:
- [ ] HttpClient shouldn't be a RequestPolicy-- perhaps should fill the role of RequestPolicyChain/RequestPolicyManager.
- [ ] Turn magic numbers into named constants
- [ ] RefreshTokenClient shouldn't use Retrofit/OkHttp

RikkiGibson:
RefreshTokenClient depends on java.net.Proxy. I want to figure out a reasonable way to support that in RxNettyAdapter before porting/breaking it off of OkHttp/Retrofit, so probably after this PR.

RikkiGibson:
Making HttpClient do the job of RequestPolicyChain will take a little thought in the design. For instance, sendRequestAsync would probably become protected and some new public entry point would be exposed to make sure requests pass through the policies when they are issued.

RikkiGibson:
@daschult if you have no objections to 704a30f I might merge this branch today.

daschult:
@RikkiGibson Looks good to me. Go ahead.

PR #216, opened 2017-09-12
Add AzureProxy with simple LRO support

daschult:
This is my first pass over adding AzureProxy (azure-client-runtime wrapper around RestProxy) and long running operation support. This pull request contains a basic implementation and tests for:
1. Synchronously passing proxy implementations through AzureProxy to RestProxy (see AzureProxyTests.syncGet),
2. Synchronously polling a long running operation that returns a Location header (see AzureProxyTests.syncCreateWithLocation),
3. Asynchronously passing proxy implementations through AzureProxy to RestProxy (see AzureProxyTests.asyncGet), and
4. Asynchronously polling a long running operation that returns a Location header (see AzureProxyTests.asyncCreateWithLocation).

This pull request also adds the concept of a ResponseHandler to RestProxy. This allows a wrapping class to insert functionality into a Swagger Interface's proxy implementation without having to make RestProxy an instance class.

I know that I probably don't handle Location how Azure would expect me to, but just look at the basic flow. MockAzureHttpClient now has the beginnings of a mock Azure Service Provider, so we can iterate pretty quickly on this.


azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    public static <A> A create(Class<A> swaggerInterface, final HttpClient httpClient, SerializerAdapter<?> serializer) {
+        return RestProxy.create(swaggerInterface, httpClient, serializer, new RestProxy.ResponseHandler() {
+            @Override
+            public Object handleSyncResponse(HttpResponse response, SwaggerMethodParser methodParser, SerializerAdapter<?> serializer) throws IOException, RestException {
+                while (response.statusCode() == 202) {
+                    final String location = response.headerValue("Location");
+                    final HttpRequest pollRequest = new HttpRequest(methodParser.fullyQualifiedMethodName(), "GET", location);
+                    response = httpClient.sendRequest(pollRequest);
```
RikkiGibson: Wait for the number of seconds denoted by the Retry-After response header.
https://github.com/Azure/autorest-clientruntime-for-java/blob/1abab34b13adf2d7dba7e5ecc73b376e6f7a113e/azure-client-runtime/src/main/java/com/microsoft/azure/PollingState.java#L426

Probably a lot of the little details can be recreated from this class, or this class will prove to still be useful after OkHttp/Retrofit has been factored out of it.

RikkiGibson:
So, all of our samples call `ResourceGroup.beginDeleteById` as part of their cleanup. It seems likely enough that we want to support that use case, which is one of the main reasons to implement LROs as a RestProxy feature rather than a RequestPolicy-- you can check an annotation value and decide whether to poll the LRO or to immediately return the fact that the LRO was initiated successfully.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -0,0 +1,112 @@
...
+                                else {
+                                    final Value<String> location = new Value<>(response.headerValue("Location"));
+                                    result = Observable.just(true)
+                                            .flatMap(new Func1<Boolean, Observable<? extends HttpResponse>>() {
+                                                @Override
+                                                public Observable<? extends HttpResponse> call(Boolean aBoolean) {
+                                                    final HttpRequest pollRequest = new HttpRequest(methodParser.fullyQualifiedMethodName(), "GET", location.get());
+                                                    return httpClient.sendRequestAsync(pollRequest).toObservable();
+                                                }
+                                            })
```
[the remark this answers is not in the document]
daschult: Good suggestion. Done.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
daschult: Added basic support for Retry-After. I think it'll be smart to go through PollingState and to create unit tests for each of the scenarios that it accounts for. Then I can easily go through AzureProxy and make sure all of the cases are covered.

daschult:
@RikkiGibson I was really hoping that we didn't have a good use case for the begin*() scenario, but it sounds like if we're using it in our samples then we should do it. After this pull request I'll add an annotation for LongRunningOperation.

RikkiGibson:
My ideal would really be to make use of the fact that Observables can emit multiple events and have one method that produces an observable that emits an event when the LRO is initiated and another event when the LRO completes. I couldn't say if that's feasible/desirable though.

RikkiGibson:
Depending on how people are using this stuff-- "I know in advance that I either don't care until it's really done, or I don't care after it's initiated"--it might be more complex for them, anyway.

daschult:
After our talk this morning, I'm going to move forward with this work. @anuchandy @RikkiGibson Thanks for your comments and design help!

PR #220, opened 2017-09-14
Long running operations

daschult:
This pull request adds support for long running operations with no return type (some delete calls are this way). It also adds support for polling long running operations and returning periodic OperationStatus objects.

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
@@ -128,20 +128,4 @@ public class AzureProxy {
 
         return retryAfterSeconds;
     }
-
-    private static class Value<T> {
-        private T value;
-
-        Value(T value) {
-            set(value);
-        }
-
-        public T get() {
-            return value;
-        }
-
-        public void set(T value) {
-            this.value = value;
-        }
-    }
 }
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
@@ -6,19 +6,24 @@
 
 package com.microsoft.azure.v2;
 
+import com.google.common.reflect.TypeToken;
 import com.microsoft.rest.RestException;
 import com.microsoft.rest.protocol.SerializerAdapter;
+import com.microsoft.rest.v2.InvalidReturnTypeException;
 import com.microsoft.rest.v2.RestProxy;
 import com.microsoft.rest.v2.SwaggerMethodParser;
 import com.microsoft.rest.v2.http.HttpClient;
 import com.microsoft.rest.v2.http.HttpRequest;
 import com.microsoft.rest.v2.http.HttpResponse;
+import rx.Completable;
 import rx.Observable;
 import rx.Single;
 import rx.functions.Func0;
 import rx.functions.Func1;
 
 import java.io.IOException;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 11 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/v2/Value.java
```diff
@@ -0,0 +1,50 @@
...
+ * license information.
+ */
+
+package com.microsoft.azure.v2;
+
+/**
+ * A container for a generic type. Serves a similar purpose as pointers in C/C++.
+ * @param <T>
+ */
+class Value<T> {
```
RikkiGibson: Maybe mention that this is a workaround for Java disallowing mutation of local variables in closures. That's basically the usage of it, right?

daschult: Yeah, that's it. I'll do that.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -115,6 +202,31 @@ static String getPollUrl(HttpResponse response, String currentPollUrl) {
...
 
+    private static Observable<HttpResponse> sendPollRequestWithDelay(SwaggerMethodParser methodParser, final HttpClient httpClient, final Value<String> pollUrl, final Value<Long> retryAfterSeconds) {
+        final String fullyQualifiedMethodName = methodParser.fullyQualifiedMethodName();
+
+        return Observable.defer(new Func0<Observable<HttpResponse>>() {
+            @Override
+            public Observable<HttpResponse> call() {
+                final HttpRequest pollRequest = createPollRequest(fullyQualifiedMethodName, pollUrl.get());
+
+                Single<HttpResponse> pollResponse = httpClient.sendRequestAsync(pollRequest);
```
anuchandy: here we are taking an assumption that the observable that `httpClient.sendRequestAsync` return is cold so that applying delay later in this observable ensure actual request is really delayed.

I think we could cover both cold and hot by applying delay operator after defer, something like:

```java
return Observable.defer(new Func0<Observable<HttpResponse>>() {
   @Override
   public Observable<HttpResponse> call() {
       final HttpRequest pollRequest = createPollRequest(fullyQualifiedMethodName, pollUrl.get());
       Single<HttpResponse> pollResponse = httpClient.sendRequestAsync(pollRequest);
       return pollResponse.toObservable();
  }
})
.delay(retryAfterSeconds.get() != null ? retryAfterSeconds.get() : 0, TimeUnit.SECONDS);
```

What do you think?

daschult: Sure, that sounds fine. I'll create a pull request for that.

daschult: I started implementing this, but then I realized that this won't work as planned. The delay call must be inside the defer() because it needs to re-evaluated for each poll request since retryAfterSeconds may have been updated since the previous poll request. If the delay() is done outside of the defer(), then the delay will be calculated only once and then will be reapplied to every poll request, but the amount of delay will never be updated, even if retryAfterSeconds changes.
In short, I think the delay needs to stay in the defer().

anuchandy: @daschult .. sorry, you are right, the current version of delay will not reevaluate the interval. 

Thinking more about this, still feel that relying on the assumption that observable always cold is wrong. What if we revert the logic, i.e. we calculate the delay inside defer and make actual http request outside.. like:

```java
Observable.defer(() -> Observable.just(true).delay(retryAfterSeconds.get() != null ? retryAfterSeconds.get() : 0, TimeUnit.SECONDS))
.flatMap(b -> {
    final HttpRequest pollRequest = createPollRequest(fullyQualifiedMethodName, pollUrl.get());
    Single<HttpResponse> pollResponse = httpClient.sendRequestAsync(pollRequest);
    return pollResponse.toObservable();
});
```

RikkiGibson: We're talking about the observable that performs a polling request, right? It's important that such an observable performs a request when it's subscribed to, each time it's subscribed to, and not at other times.

anuchandy: Which one is correct?

1. Perform request as soon as we subscribe and delay the response for the given timespan
2. delay the request for the given timespan and then send request and return response.

The expectation is - method `sendPollRequestWithDelay` is for sending a SINGLE poll request with delay. Above code basically work as expected right?.. when the caller subscribed to the downstream observable, the code inside the defer will run which produce a delay and when the delay done then the request will be send..

RikkiGibson: I think it should be 2. and it looks like you're right.

daschult: Yes, we should delay, send the request, and immediately emit/return the response. I'll see if I can come up with something for this in a pull request.

anuchandy: This should do the trick (delay before request, reevaluate delay each time subscription happens).. 

```java
Observable.defer(() -> Observable.just(true).delay(retryAfterSeconds.get() != null ? retryAfterSeconds.get() : 0, TimeUnit.SECONDS))
.flatMap(b -> {
    final HttpRequest pollRequest = createPollRequest(fullyQualifiedMethodName, pollUrl.get());
    Single<HttpResponse> pollResponse = httpClient.sendRequestAsync(pollRequest);
    return pollResponse.toObservable();
});
```

PR #224, opened 2017-09-19
Add support for Location and Azure-AsyncOperation long running operations

daschult:
This pull request adds support for long running operations that use the Location or the Azure-AsyncOperation protocols.

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
@@ -1,0 +1,160 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure.v2;
+
+import com.microsoft.rest.protocol.SerializerAdapter;
+import com.microsoft.rest.v2.http.HttpRequest;
+import com.microsoft.rest.v2.http.HttpResponse;
+import rx.Single;
+import rx.functions.Func1;
+
+import java.io.IOException;
+
+/**
+ * A PollStrategy type that uses the Azure-AsyncOperation header value to check the status of a long
+ * running operation.
+ */
+public final class AzureAsyncOperationPollStrategy extends PollStrategy {
+    private final String fullyQualifiedMethodName;
+    private final String operationResourceUrl;
+    private final String originalResourceUrl;
+    private final SerializerAdapter<?> serializer;
+    private Boolean pollingSucceeded;
+    private boolean gotResourceResponse;
+
+    /**
+     * The name of the header that indicates that a long running operation will use the
+     * Azure-AsyncOperation strategy.
+     */
+    public static final String HEADER_NAME = "Azure-AsyncOperation";
+
+    /**
+     * The provisioning state of the operation resource if the operation is still in progress.
+     */
+    public static final String IN_PROGRESS = "InProgress";
+
+    /**
+     * The provisioning state of the operation resource if the operation is successful.
+     */
+    public static final String SUCCEEDED = "Succeeded";
+
+    /**
+     * Create a new AzureAsyncOperationPollStrategy object that will poll the provided operation
+     * resource URL.
+     * @param fullyQualifiedMethodName The fully qualified name of the method that initiated the
+     *                                 long running operation.
+     * @param operationResourceUrl The URL of the operation resource this pollStrategy will poll.
+     * @param serializer The serializer that will deserialize the operation resource and the
+     *                   final operation result.
+     */
+    private AzureAsyncOperationPollStrategy(String fullyQualifiedMethodName, String operationResourceUrl, String originalResourceUrl, SerializerAdapter<?> serializer) {
+        this.fullyQualifiedMethodName = fullyQualifiedMethodName;
+        this.operationResourceUrl = operationResourceUrl;
+        this.originalResourceUrl = originalResourceUrl;
+        this.serializer = serializer;
+    }
+
+    @Override
+    public HttpRequest createPollRequest() {
+        String pollUrl = null;
+        if (pollingSucceeded == null) {
+            pollUrl = operationResourceUrl;
+        }
+        else if (pollingSucceeded) {
+            pollUrl = originalResourceUrl;
+        }
+        return new HttpRequest(fullyQualifiedMethodName, "GET", pollUrl);
+    }
+
+    @Override
+    public void updateFrom(HttpResponse httpPollResponse) throws IOException {
+        updateDelayInMillisecondsFrom(httpPollResponse);
+
+        if (pollingSucceeded == null) {
+            final String bodyString = httpPollResponse.bodyAsString();
+            updateFromResponseBodyString(bodyString);
+        }
+        else if (pollingSucceeded) {
+            gotResourceResponse = true;
+        }
+    }
+
+    @Override
+    public Single<HttpResponse> updateFromAsync(final HttpResponse httpPollResponse) {
+        Single<HttpResponse> result;
+
+        if (pollingSucceeded == null) {
+            updateDelayInMillisecondsFrom(httpPollResponse);
+
+            result = httpPollResponse.bodyAsStringAsync()
+                    .flatMap(new Func1<String, Single<HttpResponse>>() {
+                        @Override
+                        public Single<HttpResponse> call(String bodyString) {
+                            Single<HttpResponse> result = Single.just(httpPollResponse);
+                            try {
+                                updateFromResponseBodyString(bodyString);
+                            } catch (IOException e) {
+                                result = Single.error(e);
+                            }
+                            return result;
+                        }
+                    });
+        }
+        else {
+            if (pollingSucceeded) {
+                gotResourceResponse = true;
+            }
+
+            result = Single.just(httpPollResponse);
+        }
+
+        return result;
+    }
+
+    private void updateFromResponseBodyString(String httpResponseBodyString) throws IOException {
+        final OperationResource operationResource = serializer.deserialize(httpResponseBodyString, OperationResource.class);
+        if (operationResource != null) {
+            final String provisioningState = provisioningState(operationResource);
+            if (!IN_PROGRESS.equalsIgnoreCase(provisioningState)) {
+                pollingSucceeded = SUCCEEDED.equalsIgnoreCase(provisioningState);
+                clearDelayInMilliseconds();
+            }
+        }
+    }
+
+    private static String provisioningState(OperationResource operationResource) {
+        String provisioningState = null;
+
+        final OperationResource.Properties properties = operationResource.properties();
+        if (properties != null) {
+            provisioningState = properties.provisioningState();
+        }
+
+        return provisioningState;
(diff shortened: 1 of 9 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/v2/OperationStatus.java
```diff
@@ -6,54 +6,134 @@
...
 /**
- * The status of a long running operation. This generally is created from the result of polling for
- * whether a long running operation is done or not.
- * @param <T>
+ * The current state of polling for the result of a long running operation.
  */
-public final class OperationStatus<T> {
-    private final boolean isDone;
-    private final T result;
+class OperationStatus<T> {
```
RikkiGibson: Is this no longer something you want referenced in Azure gen code?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Update the delay in milliseconds from the provided HTTP poll response.
+     * @param httpPollResponse The HTTP poll response to update the delay in milliseconds from.
+     */
+    final void updateDelayInMillisecondsFrom(HttpResponse httpPollResponse) {
+        final String retryAfterSecondsString = httpPollResponse.headerValue("Retry-After");
+        if (retryAfterSecondsString != null && !retryAfterSecondsString.isEmpty()) {
+            try {
+                final long responseDelayInMilliseconds = Long.valueOf(retryAfterSecondsString) * 1000;
+                delayInMilliseconds = Math.max(responseDelayInMilliseconds, AzureProxy.defaultDelayInMilliseconds());
```
anuchandy: why we take max out of these two? Should we use `AzureProxy.defaultDelayInMilliseconds()` as delay value only if "Retry-After" header not present in the response?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+            catch (NumberFormatException ignored) {
```
anuchandy: In case parsing of `Retry-After` header fail should we fall back to `AzureProxy.defaultDelayInMilliseconds()`?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Update the status of this PollStrategy from the provided HTTP poll response.
+     * @param httpPollResponse The response of the most recent poll request.
+     * @return A Completable that can be used to chain off of this operation.
+     */
+    abstract Single<HttpResponse> updateFromAsync(HttpResponse httpPollResponse);
+
+    /**
+     * Get whether or not this PollStrategy's long running operation is done.
+     * @return Whether or not this PollStrategy's long running operation is done.
```
anuchandy: in general when the method description and return value description are same we just remove the method description and keep the other one.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
```diff
@@ -0,0 +1,160 @@
...
+            gotResourceResponse = true;
+        }
+    }
+
+    @Override
+    public Single<HttpResponse> updateFromAsync(final HttpResponse httpPollResponse) {
+        Single<HttpResponse> result;
+
+        if (pollingSucceeded == null) {
+            updateDelayInMillisecondsFrom(httpPollResponse);
```
anuchandy: in `updateFrom` sync method `updateDelayInMillisecondsFrom` is always called, but in async we do that only if `pollingSucceeded == null`, which one is correct?

anuchandy: should we implement sync `updateFrom` using `updateFromAsync` (by blocking on it)?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -161,21 +172,13 @@ else if (returnTypeToken.isSubtypeOf(Observable.class)) {
...
-                                                        // marked as done, so we set it here and then concatWith() it
-                                                        // later to force the Observable to return the operationStatus
-                                                        // that is done.
-                                                        lastOperationStatus.set(operationStatus);
-                                                    }
-                                                    return stop;
+                                                    return operationStatus.isDone();
                                                 }
                                             })
                                             .concatWith(Observable.defer(new Func0<Observable<OperationStatus<Object>>>() {
```
anuchandy: Interesting.. as per doc [takeUntil](http://reactivex.io/documentation/operators/images/takeUntil.p.png) is suppose to forward the event meeting the criteria as final event

```java
final int[] i = { 0 };
Observable.defer(() -> Observable.just(i[0]++))
            .repeat()
            .takeUntil(v -> v == 5)
            .subscribe(System.out::println);
```

output is

```
0
1
2
3
4
5
```

Anyway, if we find that concat is needed can we add some comments (seems we already had some note but it got deleted in this pr)?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/OperationStatus.java
daschult: We definitely want this referenced in gen code. Fixed.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
daschult: I was thinking of the scenario where a user sets the AzureProxy.defaultDelayInMilliseconds to 30 seconds, but the Retry-After header says 3 seconds. In that case, the user is in no hurry to poll again, so we can wait until the user specified 30 seconds. However, if the user specifies 30 seconds but the Retry-After header says 60 seconds, then we should respect Azure's wishes and not poll for another 60 seconds. Does this makes sense? Or should we always respect Azure's wishes?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * A PollStrategy type that uses the Azure-AsyncOperation header value to check the status of a long
+ * running operation.
+ */
+public final class AzureAsyncOperationPollStrategy extends PollStrategy {
+    private final String fullyQualifiedMethodName;
+    private final String operationResourceUrl;
+    private final String originalResourceUrl;
+    private final SerializerAdapter<?> serializer;
+    private Boolean pollingSucceeded;
```
anuchandy: there are few methods in this class which checks `pollingSucceeded` is NULL vs TRUE vs FALSE. I wish there Nullable<T> in jdk.. :) ..

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
anuchandy: ah.. got it now, didn't know that user can configure `defaultDelayInMilliseconds`. Now the logic make sense. If user don't set it then library uses a default value - right?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
RikkiGibson: Even `Boolean.isTrue()` which doesn't require a null comparison would be welcome.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
daschult: Correct.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
daschult: I've changed this slightly so that PollStrategy gets passed an initial delayInMilliseconds value in its constructor. If Azure never sends a Retry-After header, then the PollStrategy will continue to use the initial delayInMilliseconds as its delay duration. If Azure ever sends a Retry-After header, then that will become the new delayInMilliseconds value and we'll forget about the previous delayInMilliseconds value. I think this is simpler and it should work fine.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
daschult: Removed @return.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
```diff
@@ -0,0 +1,160 @@
...
+     * The provisioning state of the operation resource if the operation is successful.
+     */
+    public static final String SUCCEEDED = "Succeeded";
+
+    /**
+     * Create a new AzureAsyncOperationPollStrategy object that will poll the provided operation
+     * resource URL.
+     * @param fullyQualifiedMethodName The fully qualified name of the method that initiated the
+     *                                 long running operation.
+     * @param operationResourceUrl The URL of the operation resource this pollStrategy will poll.
```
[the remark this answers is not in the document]
daschult: Added.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
anuchandy: sounds good!

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
daschult: I didn't want to do that originally because of the comments that Jeffrey Richter made about synchronous code going through asynchronous code paths adds a lot of overhead. However, I think coding to avoid that right now adds a fair amount of maintenance (from the duplication). I'll Change updateFrom() to call updateFromAsync().

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
daschult: That's bizarre. I just removed the concatWith() call and all of my tests pass. I remember needing to put that in, but I must've been using a different operator and switched it to takeUntil(). Weird. I've removed the concatWith() now.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
daschult: I was using the pollingSucceeded Boolean as a ternary state value (null meant that polling hadn't completed, true meant that it has completed successfully, false meant it had completed unsuccessfully), but I've converted it into two booleans (pollingCompleted and pollingSucceeded).

PR #225, opened 2017-09-19
Add support for AzureHost annotation

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerInterfaceParser.java
```diff
@@ -30,14 +30,17 @@
...
-        }
-        else if (host != null) {
+        if (host != null && !host.isEmpty()) {
             this.host = host;
-        } else {
-            this.host = hostAnnotation.value();
+        }
+        else {
+            final Host hostAnnotation = swaggerInterface.getAnnotation(Host.class);
+            if (hostAnnotation != null && !hostAnnotation.value().isEmpty()) {
```
anuchandy: any edge case where `hostAnnotation.value()` may be null hence expression `hostAnnotation.value().isEmpty()` throws NPE?

RikkiGibson: If someone decorates their interface with this annotation and passes null to it... well, I just don't think they're doing themselves any favors.

RikkiGibson:
This work raises the question of how to provide the AzureEnvironment object to a generated service client.

Right now, we mainly have methods in e.g. StorageManager which instantiate AzureTokenCredentials, get the environment from those credentials, and apply the base URL for the service to the RestClient object.

With an approach like this, it's necessary to somehow provide an AzureEnvironment to the generated client, e.g. StorageManagementClientImpl.
- Adding an Azure-specific property to RestClient is difficult because RestClient doesn't know about azure-client-runtime data types.
- Extending RestClient is difficult because it means there are parameters like baseURL which we intend to ignore for our more Azure-specific method of resolving a base URL, as well as the fact that you'll need to rewrite the builder.
- Creating a new, separate data type, e.g. AzureConfig, which would be like an Azure-specific RestClient, is a tough sell as well because there's a lot of code to churn to get the name RestClient out of all the generated as well as handwritten code.

In short, I'm starting to lean toward thinking `@AzureHost` needs to provide significant value to justify the changes needed for it. So do we feel that it does and why?

RikkiGibson:
At this moment I think the problem is most simply solved by adding the baseURL parameter to `AzureProxy.create()` and relying on the existing infrastructure, but if anyone has any concerns or other perspectives to share I'd be happy to discuss.

daschult:
I'll add an overload to AzureProxy.create() that takes a baseUrl argument instead of the AzureEnvironment. I think we should prefer to use the AzureEnvironment argument overload, but I'm okay with adding the baseUrl overload to make the transition easier.

PR #226, opened 2017-09-21
Response logging and LRO helpers

RikkiGibson:
This PR modifies HttpResponse to add a `headers()` property to support logging all the headers on a response. It includes some preliminary work for response logging--it may need to be tweaked to determine what kinds of response bodies should be logged, etc.

There are also some changes to some ServiceFuture factory methods to support runtime changes, mainly related to LROs. I added the ServiceFutureUtil class because I wanted a factory method that knew about some Azure stuff, and I suspect that AzureServiceFuture is not going to be of use to us pretty soon because we're deprecating the callback-based API for paging.


azure-client-runtime/src/main/java/com/microsoft/azure/v2/util/ServiceFutureUtil.java
```diff
@@ -0,0 +1,21 @@
...
+import rx.Observable;
+import rx.Single;
+import rx.functions.Func1;
+
+public class ServiceFutureUtil {
+    public static <T> ServiceFuture<T> fromLRO(Observable<OperationStatus<T>> observable, ServiceCallback<T> callback) {
+        Single<T> single = observable.last().toSingle().map(new Func1<OperationStatus<T>, T>() {
+            @Override
+            public T call(OperationStatus<T> tOperationStatus) {
+                return tOperationStatus.result();
```
daschult: This assumes that the operation completed successfully. Is that okay or is this only temporary?

RikkiGibson: That's a good question. I think it depends on OperationStatus.result throwing if an error occurred, which might not be what we want.

anuchandy: if `tOperationStatus.result()` throws exception the map operator will catch it and invoke `onError` in the downstream, which get translated to a call to `serviceFuture.setException(throwable)`.

If there is an error occurred during polling then `Observable<OperationStatus<T>>` should call `orError` on the downstream, which in this case again will get propagated and translated to a call to `serviceFuture.setException(throwable)`.

If above assumptions are correct then this implementation is correct, right?

daschult: Yes, the implementation is correct.

RikkiGibson: We determined that it's acceptable to assume that either the last OperationStatus should contain a non-null `.result()` or that something should have thrown somewhere earlier on causing Single.error to be returned to whoever finally uses this Single.

Basically, this code is OK, except the parameter name in the Func1 is wacky.

PR #229, opened 2017-09-22
Remove v2 packages

daschult:
Removes the v2 packages and moves their contents down into the main packages.


client-runtime/src/main/java/com/microsoft/rest/interceptors/BaseUrlHandler.java
```diff
@@ -1,53 +0,0 @@
...
-
-import java.io.IOException;
-
-/**
- * Handles dynamic replacements on base URL. The arguments must be in pairs
- * with the string in raw URL to replace as replacements[i] and the dynamic
- * part as replacements[i+1]. E.g. {subdomain}.microsoft.com can be set
- * dynamically by setting header x-ms-parameterized-host: "{subdomain}, azure"
- */
-public final class BaseUrlHandler implements Interceptor {
```
RikkiGibson: This is now handled by RestProxy, right?

daschult: Correct.

jianghaolu:
We will need to keep v2 eventually - so that people can use v1 and v2 side by side.

daschult:
What's the scenario that someone would use v1 and v2 side by side?

RikkiGibson:
@daschult It's needed if someone is using both azure-sdk-for-java and, say, azure-keyvault-java or any of the other standalone SDKs that hasn't updated to the new runtime yet. It seems that other widely depended on packages like RxJava and OkHttp use a similar convention for dealing with this, which we've seen in all the `import okhttp3.something` import statements in our code.

PR #231, opened 2017-09-23
Support byte[] body params

RikkiGibson:
This simply makes it so when a byte[] is passed as a `@BodyParam`, RestProxy uses it as the request body. There are also some small changes that prevent setting multiple Content-Length header values, respect the Content-Type set in `@Headers`, and provide robustness when a server neglects to pass a Content-Length header on a response.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
@@ -131,9 +131,17 @@ public class RestProxy implements InvocationHandler {
             request.withHeader(header.name(), header.value());
         }
 
+        String mimeType = request.headers().value("Content-Type");
         final Object bodyContentObject = methodParser.body(args);
-        if (bodyContentObject != null) {
-            final String mimeType = "application/json";
+        if (bodyContentObject instanceof byte[]) {
+            if (mimeType == null) {
+                mimeType = "application/octet-stream";
+            }
+            request.withBody((byte[]) bodyContentObject, mimeType);
+        } else if (bodyContentObject != null) {
+            if (mimeType == null) {
+                mimeType = "application/json";
+            }
             final String bodyContentString = serializer.serialize(bodyContentObject);
             request.withBody(bodyContentString, mimeType);
         }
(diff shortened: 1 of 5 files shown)


client-runtime/src/main/java/com/microsoft/rest/http/HttpRequest.java
```diff
@@ -103,7 +103,8 @@ public HttpRequest withBody(byte[] body, String mimeType) {
     public HttpRequest withBody(HttpRequestBody body, String mimeType) {
         this.body = body;
         this.mimeType = mimeType;
-        return withHeader("Content-Length", String.valueOf(body.contentLength()));
+        headers.set("Content-Length", String.valueOf(body.contentLength()));
```
daschult: What was wrong with "withHeader()"?

client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -131,9 +131,17 @@ private HttpRequest createHttpRequest(SwaggerMethodParser methodParser, Object[]
             request.withHeader(header.name(), header.value());
         }
 
+        String mimeType = request.headers().value("Content-Type");
```
daschult: I think we should talk about this line. This says to me that we're allowing our service/swagger interfaces (I think I'm just going to call them service interfaces from now on. That's what we call them in our code) to set Content-Type as a header. The problem I have with this is that Content-Type is tied very closely to the body, and it's kind of a big deal. If the Content-Type is set incorrectly, then the target server might not function correctly (the same goes for the "Accept" header, but that's a different discussion). Since we're the client libraries, I'm leaning towards us always setting the Content-Type header based on the type of the body argument passed to the function.
So, in short, I agree with the changes you've made on lines 136 through 144, but I'm not sure how I feel about this line. Let's leave it here for now, though.

RikkiGibson: Yeah, I think we should drill down to the root of how/why `@Headers(“Content-Type: ...”)` is being generated by AutoRest and work back to something reasonable.

client-runtime/src/main/java/com/microsoft/rest/http/HttpRequest.java
RikkiGibson: I believe it was that if Content-Length is already set, it needs to be replaced and not appended to.

RikkiGibson:
Added a test which PUTs a byte[] in httpbin.

PR #233, opened 2017-09-25
Remove ok http

daschult:
Remove OkHttp from autorest-clientruntime-for-java. This pull request adds more tests and adds the stubs for interacting with java.net.Proxy objects. Mostly, though, it just removes all references to OkHttp and Retrofit.


azure-client-runtime/src/main/java/com/microsoft/azure/util/ServiceFutureUtil.java
```diff
@@ -1,37 +0,0 @@
...
- */
-public class ServiceFutureUtil {
-    /**
-     * Creates a ServiceFuture from an observable representing a long-running operation.
-     * @param observable The observable representing the long-running operation.
-     * @param callback The callback to run when the service future is completed.
-     * @param <T> The type of entity expected to be returned from the long-running operation.
-     * @return A ServiceFuture representing the long-running operation.
-     */
-    public static <T> ServiceFuture<T> fromLRO(Observable<OperationStatus<T>> observable, ServiceCallback<T> callback) {
```
RikkiGibson: I wrote this for use in gen code for LROs-- implementing e.g. `create(ServiceCallback)` in terms of `beginCreate()`. Is there something you wanted that to migrate to?

azure-client-runtime/src/test/java/com/microsoft/azure/AzureProxyToRestProxyTests.java
```diff
@@ -597,10 +599,66 @@ public void AsyncHttpsHeadersRequest() {
...
+            service.get();
+            fail("Expected exception.");
+        }
+        catch (InvalidReturnTypeException e) {
+            assertContains(e.getMessage(), "rx.Observable<com.microsoft.azure.HttpBinJSON>");
+            assertContains(e.getMessage(), "AzureProxyToRestProxyTests$Service15.get()");
+        }
+    }
+
+    @Host("http://httpbin.org")
```
RikkiGibson: Oops, should've noticed you wrote this before I wrote a very similar thing.

client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -133,9 +134,21 @@ private HttpRequest createHttpRequest(SwaggerMethodParser methodParser, Object[]
...
+            final String contentType = methodParser.bodyContentType();
+            if (ContentType.APPLICATION_JSON.equalsIgnoreCase(contentType)) {
+                final String bodyContentString = serializer.serialize(bodyContentObject);
+                request.withBody(bodyContentString, contentType);
+            }
+            else if (bodyContentObject instanceof byte[]) {
+                request.withBody((byte[]) bodyContentObject, contentType == null ? ContentType.APPLICATION_OCTET_STREAM : contentType);
+            }
+            else if (bodyContentObject instanceof String) {
+                request.withBody((String) bodyContentObject, contentType == null ? ContentType.APPLICATION_OCTET_STREAM : contentType);
```
RikkiGibson: Is this the right content type for a string body?

client-runtime/src/main/java/com/microsoft/rest/http/ContentType.java
```diff
@@ -0,0 +1,22 @@
...
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.rest.http;
+
+/**
+ * The different values that we commonly use for Content-Type headers.
+ */
+public abstract class ContentType {
```
RikkiGibson: If this class is intended to not be instantiated or subclassed I'd prefer it be final with a private empty constructor. It might be nice to make it an enum if we can swing it at all the usage sites. That might not be very straightforward though.

client-runtime/src/main/java/com/microsoft/rest/http/RxNettyAdapter.java
```diff
@@ -62,6 +63,11 @@ private SSLEngine getSSLEngine(String host) {
     }
 
     @Override
+    public HttpClient withProxy(Proxy proxy) {
```
RikkiGibson: It looks like this is starting down a road of allowing mutation of an HttpClient instance after it's created, which gives me pause. If you find it's significantly easier to work with than an immutable class/class hierarchy, I can deal with it.

RikkiGibson: Basically, I aspired to make RestClient and its contents immutable so that you can share it with whatever other services impls or clients (e.g. RefreshTokenClient) that need it to configure itself without concern that it might mess up the configuration of the original service.

azure-client-runtime/src/main/java/com/microsoft/azure/util/ServiceFutureUtil.java
daschult: I was thinking that this Util class was only temporary until we removed the create(ServiceCallback) function. Since we're removing that function for v2 (right?), I figured that we might as well start removing it now. What kind of changes need to happen in gen code (branch v2) to not need this Util class?

azure-client-runtime/src/test/java/com/microsoft/azure/AzureProxyToRestProxyTests.java
daschult: We were writing it at the same time. I'll merge your changes into this pull request.

client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
daschult: I originally had the Content-Type set to "application/json", but that always surrounded the body with double quotes (hello world -> "hello world"). In JSON, that is the correct thing to do, but if we're just sending a String body, then I don't think we'd need to quote it. The real test is if we have any service interfaces that use String bodies. I'm guessing most of them use provider-specific objects, so we may end up removing this branch of the if-statement.

azure-client-runtime/src/main/java/com/microsoft/azure/util/ServiceFutureUtil.java
RikkiGibson: If we actually are getting rid of ServiceCallback in general then we don't need ServiceFutureUtil.fromLRO. So far I think we've just decided that ListOperationCallback is going away. I'd be open to killing ServiceCallback but I'm not sure I can make an informed decision about it.

client-runtime/src/main/java/com/microsoft/rest/http/ContentType.java
daschult: I agree about making it final with a private constructor. I went back and forth on making this an enum or just leaving it as a collection of constant strings. Let's leave it as strings for now with the option to make it an enum later.

daschult: Done.

client-runtime/src/main/java/com/microsoft/rest/http/RxNettyAdapter.java
daschult: This is a really good point. I'll change this to go on the Builder instead of the RestClient.

daschult: Scratch that. I'll make this go on the constructor of our HttpClients.

daschult: Scratch that. How an HttpClient uses a Proxy is pretty implementation specific, so I created an RxNettyAdapter Builder that has a ```withProxy(Proxy)``` method on it.

PR #237, opened 2017-09-27
Handle unexpected response status in async methods

RikkiGibson:
I wound up factoring out a few helpers and doing some slightly odd-looking things to make this work.

A particular difficult point was to implement handleSyncHttpResponse in terms of handleAsyncHttpResponse, because handleAsyncHttpResponse can return either Completable or Single, and handleAsyncHttpResponse makes incompatible assumptions about the return type of the Swagger method. I expect a little revision and maybe some more test coverage to be necessary before merge.


RikkiGibson:
It looks like some tests failed in CI, but they're passing on my local machine. I can't restart the build, and I'm not sure what the failure is about.

daschult:
Those failures are about the request/response body is a String, should it be quoted or not? I felt that it shouldn't be quoted, so I made some tests to enforce that.

RikkiGibson:
I was able to repro the error locally after merging. I just can't see what changed about the way I get the response content and deserialize it that would've caused the quotes go to away.

This test is deserializing to HttpBinJSON and printing the .data property.

client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -165,84 +166,99 @@ protected Object handleSyncHttpResponse(HttpRequest httpRequest, HttpResponse ht
         return handleSyncHttpResponse(httpRequest, httpResponse, methodParser, returnType);
     }
 
-    protected Object handleSyncHttpResponse(HttpRequest httpRequest, HttpResponse httpResponse, SwaggerMethodParser methodParser, Type returnType) throws IOException {
-        Object result;
+    protected Object handleSyncHttpResponse(HttpRequest httpRequest, HttpResponse httpResponse, SwaggerMethodParser methodParser, final Type returnType) throws IOException {
```
daschult: `returnType` doesn't need to be `final` here anymore.

PR #238, opened 2017-09-28
Add ServiceFuture back for now

RikkiGibson:
We're still deciding on what the future of our futures should be, so I'm adding this type back to prevent the need for lots of AutoRest generator changes in the meantime.


RikkiGibson:
If you have no objections I'll go ahead and merge this @daschult

PR #239, opened 2017-09-28
Make ExpectedResponses annotation not required

daschult:


RikkiGibson:
Hold up one moment on this. I'm looking at the diff in AutoRest tests and making sure this makes sense.

PR #240, opened 2017-09-28
No null or empty http headers

daschult:


client-runtime/src/main/java/com/microsoft/rest/SwaggerMethodParser.java
```diff
@@ -361,8 +361,8 @@ private static String applySubstitutions(String originalValue, Iterable<Substitu
                 if (0 <= substitutionParameterIndex && substitutionParameterIndex < methodArguments.length) {
                     final Object methodArgument = methodArguments[substitutionParameterIndex];
 
-                    String substitutionValue = String.valueOf(methodArgument);
-                    if (substitution.shouldEncode()) {
+                    String substitutionValue = methodArgument == null ? "" : methodArgument.toString();
```
RikkiGibson: So in general we don't consider it reasonable to say `@HeaderParam MyClass myClass`, right? It would be too weird to just mash JSON into a header value.

PR #242, opened 2017-09-29
Fix content type

daschult:


client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -136,25 +136,38 @@ private HttpRequest createHttpRequest(SwaggerMethodParser methodParser, Object[]
...
+                    contentType = ContentType.APPLICATION_OCTET_STREAM;
+                }
+                else {
+                    contentType = ContentType.APPLICATION_JSON;
+                }
+            }
+
+            request.headers().set("Content-Type", contentType);
+
             if (ContentType.APPLICATION_JSON.equalsIgnoreCase(contentType)) {
```
RikkiGibson: contentType will often be in the form of `application/json; charset=utf-8` so a .contains() test may be preferable

client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -136,25 +136,38 @@ private HttpRequest createHttpRequest(SwaggerMethodParser methodParser, Object[]
...
             String contentType = methodParser.bodyContentType();
+            if (contentType == null || contentType.isEmpty()) {
+                contentType = request.headers().value("Content-Type");
+            }
+            if (contentType == null || contentType.isEmpty()) {
+                if (bodyContentObject instanceof byte[] || bodyContentObject instanceof String) {
+                    contentType = ContentType.APPLICATION_OCTET_STREAM;
+                }
+                else {
+                    contentType = ContentType.APPLICATION_JSON;
```
RikkiGibson: Down the line we're going to hit funny stuff related to XML support in blob storage API (as far as I understand). C# people recently added this onto their runtime. Probably no need to address it in this PR but something to be aware of. May even be necessary to set a different serializer adapter in gen code upon detecting some attribute in the service.

client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -136,25 +136,38 @@ private HttpRequest createHttpRequest(SwaggerMethodParser methodParser, Object[]
...
+                final String bodyContentString = (String)bodyContentObject;
+                if (!bodyContentString.isEmpty()) {
+                    request.withBody((String) bodyContentObject, contentType);
                 }
-                request.withBody((String) bodyContentObject, contentType);
             }
             else {
                 final String bodyContentString = serializer.serialize(bodyContentObject);
-                request.withBody(bodyContentString, contentType);
+                if (!bodyContentString.isEmpty()) {
```
RikkiGibson: If you are able to get the serializer to produce an empty string, let me know, because it can induce some shaky server-side behaviors when an empty/null body is interpreted as JSON. Even a null `bodyContentObject` should be turned into the string `null` by the serializer.

Taking the browser's JSON serializer as a reference implementation:
```javascript
> JSON.stringify('')
< "\"\""
> JSON.stringify(null)
< "null"
```

RikkiGibson: Also:
```javascript
> JSON.parse('')
< SyntaxError: JSON Parse error: Unexpected EOF
```
Some server-side body parsers will interpret it as an empty object, but they reasonably have the right to get upset about being asked to parse an empty body into JSON. The language of "parse an empty string" versus "parse a string containing a pair of quotes" is hard to keep straight in this discussion..

client-runtime/src/main/java/com/microsoft/rest/annotations/BodyParam.java
```diff
@@ -36,5 +36,5 @@
     /**
      * @return the Content-Type that the body should be treated as.
      */
-    String contentType() default "";
+    String value() default "";
```
RikkiGibson: :+1:

PR #243, opened 2017-10-02
Throw for null body param

RikkiGibson:
This ensures that the runtime throws IllegalArgumentException when null is passed as the argument for a `@BodyParam`. It also ensure that when null is provided for a query parameter, the entire name/value pair is omitted, and if the empty string is passed, it's rendered like a non-empty string would be, like `?param=&nextparam=something`.

### DIFF
--- a/client-runtime/src/test/java/com/microsoft/rest/RestProxyTests.java
+++ b/client-runtime/src/test/java/com/microsoft/rest/RestProxyTests.java
@@ -393,6 +393,12 @@ public abstract class RestProxyTests {
         assertEquals("I'm a post body!", (String)json.data);
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void SyncPostRequestWithNullBody() {
+        final HttpBinJSON json = createService(Service8.class)
+                .post(null);
+    }
+
     @Host("http://httpbin.org")
     private interface Service9 {
         @PUT("put")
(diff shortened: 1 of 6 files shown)


RikkiGibson:
This is winding up being kind of a grab bag of small runtime tweaks that fix AutoRest tests. Maybe could use some RestProxy coverage.

client-runtime/src/test/java/com/microsoft/rest/RestProxyTests.java
```diff
@@ -394,6 +394,12 @@ public void AsyncPostRequestWithStringBody() {
         assertEquals("I'm a post body!", (String)json.data);
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void SyncPostRequestWithNullBody() {
+        final HttpBinJSON json = createService(Service8.class)
```
daschult: Remove the `final HttpBindJSON json` declaration, please. It's dead code.

client-runtime/src/test/java/com/microsoft/rest/RestProxyTests.java
```diff
@@ -879,11 +885,10 @@ public void service18GetStatus500WithExpectedResponse500() {
         HttpBinJSON putWithBodyParamApplicationOctetStreamContentTypeAndByteArrayBody(@BodyParam(ContentType.APPLICATION_OCTET_STREAM) byte[] body);
     }
 
-    @Test
+    @Test(expected = IllegalArgumentException.class)
     public void service19PutWithNoContentTypeAndStringBodyWithNullBody() {
         final HttpBinJSON result = createService(Service19.class)
```
daschult: Same here. Remove `result`.

PR #244, opened 2017-10-02
Paging support

RikkiGibson:
This PR adds support for paging. Cookies are used in the AutoRest tests to track "where" we are in a scenario that takes place over multiple requests, so I added cookie support in RxNettyAdapter.

I also added handling for some of the complex partial URLs that can be returned by services to indicate where the next page is located.


RikkiGibson:
Some of the new URL builder behaviors could use test coverage

RikkiGibson:
Will want to go back to this and pull cookies into a RequestPolicy-- the lack of cookie information in logs is already causing confusion.

Note to self: no inserting request headers in HttpClient implementations!

PR #245, opened 2017-10-03
[ADLA - ADLS] - [CODE] - Adding Data Lake resource id

ro-joowan:


ro-joowan:
There were two style errors:

[ERROR] src/main/java/com/microsoft/azure/credentials/AzureTokenCredentials.java:[53,92] (whitespace) OperatorWrap: '||' should be on a new line.
[ERROR] src/main/java/com/microsoft/azure/AzureEnvironment.java:[214,48] (blocks) LeftCurly: '{' at column 48 should have line break after.

I addressed the first one in my second commit.  The second error, however, I will leave for now in order to follow the current pattern in that file.

azure-client-runtime/src/main/java/com/microsoft/azure/AzureEnvironment.java
```diff
@@ -202,6 +209,11 @@ public String graphEndpoint() {
     }
 
     /**
+     * @return the Data Lake resource ID.
+     */
+    public String dataLakeEndpointResourceId() { return endpoints.get("dataLakeEndpointResourceId"); }
```
jianghaolu: [ERROR] src/main/java/com/microsoft/azure/AzureEnvironment.java:[214,48] (blocks) LeftCurly: '{' at column 48 should have line break after.

PR #247, opened 2017-10-03
Make OperationStatus immutable

daschult:


azure-client-runtime/src/main/java/com/microsoft/azure/AzureProxy.java
```diff
@@ -133,19 +135,19 @@ protected Object handleAsyncHttpResponse(final HttpRequest httpRequest, Single<H
                     .flatMap(new Func1<HttpResponse, Single<? extends HttpResponse>>() {
                         @Override
                         public Single<? extends HttpResponse> call(HttpResponse httpResponse) {
-                            final OperationStatus<Object> operationStatus = new OperationStatus<>(httpRequest, httpResponse, serializer);
+                            final PollStrategy pollStrategy = createPollStrategy(httpRequest, httpResponse, serializer);
 
                             Single<HttpResponse> result;
-                            if (operationStatus.isDone()) {
+                            if (pollStrategy == null || pollStrategy.isDone()) {
```
RikkiGibson: a null PollStrategy means this is not a long-running operation.

daschult: Correct.

azure-client-runtime/src/main/java/com/microsoft/azure/OperationStatus.java
```diff
@@ -23,102 +15,28 @@
...
-    /**
-     * Update the properties of this OperationStatus from the provided response.
-     * @param httpResponse The HttpResponse from the most recent request.
+     * Create a new OperationStatus with the provided PollStrategy.
+     * @param pollStrategy The polling strategy that the OperationStatus will use to check the
+     *                     progress of a long running operation.
      */
-    void updateFrom(HttpResponse httpResponse) throws IOException {
-        pollStrategy.updateFrom(httpResponse);
+    OperationStatus(PollStrategy pollStrategy) {
```
RikkiGibson: How do consumers of this value check the ProvisioningState? I'm looking around the management SDK and I'm finding that many services have their own independent ProvisioningState enum with an associated String value.

When a user receives an OperationStatus value they may want to be able to see e.g. I'm in the CREATING ProvisioningState versus the RESOLVING_DNS ProvisioningState.

daschult: Sure, that makes sense. I'll create another pull request where I add in a provisioningState() getter.

PR #253, opened 2017-10-06
Azure management support

RikkiGibson:
This features a few preliminary changes to the runtime in order to support the management SDK in practice. With these changes, I have our storage accounts and usages tests working :tada:, which is basically all the tests I saw for storage management.

- Put LoggingPolicy at the end of the chain because it lets you see everything that's been done to the HttpRequest.
- Added a Proxy field because AzureConfigurableImpl wanted to be able to set one and our concrete HTTP client can now use one.
- Refactored credentials to reflect that `authorizationHeaderValue` may throw.  Tweaked things so that AzureTokenCredentials will actually try to get a token. Not sure how this stuff can be tested except to run tests in management SDK.

I also merged master in, so some changes e.g. MSICredentials and FlatteningSerializer are just along for the ride.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/RestClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/RestClient.java
@@ -20,6 +20,7 @@ import com.microsoft.rest.http.HttpClient;
 import com.microsoft.rest.http.RxNettyAdapter;
 import com.microsoft.rest.serializer.JacksonAdapter;
 
+import java.net.Proxy;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


client-runtime/src/main/java/com/microsoft/rest/RestClient.java
```diff
@@ -41,6 +43,7 @@
     private final List<RequestPolicy.Factory> customPolicyFactories;
 
     private RestClient(RestClient.Builder builder) {
+        this.proxy = builder.proxy;
```
daschult: Where do you actually use the proxy within the RxNettyAdapter?

RikkiGibson: About to fix this. `httpClientBuilder.withProxy()` adds a Netty handler for dealing with the proxy. Would be nice to devise some test case that proves that the proxy works.

PR #254, opened 2017-10-06
Add HTTP client factory

RikkiGibson:
When working on tests for management SDK, which depend heavily on recording and playing back HTTP traffic, I found a need to specify a mock HTTP client in RestClient. RestClient had a hard reference on RxNettyAdapter so I needed to rework things a bit.

After some experimentation, I settled on this as a pretty good way to create an HTTP client from some commonly-known configuration set. It concerns me a bit that it's, well, factories and builders all the way down, but it seems to be the cost of immutability in Java 7. One alternative would be to take some of these named Factory interfaces and simply use `Func1<Configuration, Result>`. See what you think.

I also think RestClient.Builder deserves a little thought on which parameters are required-- I'd tend to advocate making those required parameters (the ones that throw if you build without setting them) be arguments to the builder's constructor as well as having `.with` methods, but it would break a lot of usage sites. In the long run, it would provide the advantage of pushing runtime errors back to compile time, which should make RestClient's users happier.

I'm also finding a lot of places where RequestPolicies need to read response body content. So far I think it is working more smoothly to have this BufferedHttpResponse instead of asking the concrete HTTP clients to cache their response body content. The caveat is that any RequestPolicy that reads response body content needs to `.buffer()` and return the BufferedHttpResponse. I'm not married to it working this way though, so feel free to play with it.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/RestClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/RestClient.java
@@ -30,6 +30,7 @@ import java.util.concurrent.TimeUnit;
  * An instance of this class stores configuration for setting up specific service clients.
  */
 public final class RestClient {
+    private final HttpClient.Factory httpClientFactory;
     private final HttpClient httpClient;
     private final Proxy proxy;
     private final String baseURL;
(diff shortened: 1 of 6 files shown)


client-runtime/src/main/java/com/microsoft/rest/RestClient.java
```diff
@@ -194,7 +196,19 @@ private Builder(final RestClient restClient) {
         /**
          * Creates an instance of the builder.
          */
-        public Builder() { }
+        public Builder() {
+            this.httpClientFactory = new RxNettyAdapter.Factory();
```
RikkiGibson: This line concerns me. I want to have a reasonable way of adding the required parameters to a builder, and a reasonable way of referencing the concrete dependencies.

It might wind up being best to just create this default factory when building if the given factory is null.

RikkiGibson:
@daschult If you are blocked on BufferedHttpResponse feel free to merge d4fbe9c into your branch at any point. Going to try and come fresh at the immutable data problem this afternoon.

RikkiGibson:
I'm opting not to go for the refactor pushing RestClient fields into an HttpClientBuilder for now. I've been thinking a lot about how to do sound, convenient immutable data in Java with composition, round tripping, and properties computed based on other properties, but I'm not ready to conclude on it and go through the code churn. For now I'm satisfied with leaving the plain old data object with configuration values and the abstract class with behavior that gets created from that plain old data separate.

PR #256, opened 2017-10-09
Add provisioning state poll strategy

daschult:
Add the missing polling strategy (ProvisioningState). This pull request also adds a CompletedPollStrategy, which takes the place of the previous null PollStrategy.

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/AzureProxy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/AzureProxy.java
@@ -8,6 +8,7 @@ package com.microsoft.azure;
 
 import com.google.common.reflect.TypeToken;
 import com.microsoft.azure.annotations.AzureHost;
+import com.microsoft.rest.http.BufferedHttpResponse;
 import com.microsoft.rest.protocol.SerializerAdapter;
 import com.microsoft.rest.InvalidReturnTypeException;
 import com.microsoft.rest.RestProxy;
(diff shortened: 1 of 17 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/AzureProxy.java
```diff
@@ -117,144 +118,95 @@ protected Object handleAsyncHttpResponse(final HttpRequest httpRequest, Single<H
...
+                            if (originalHttpRequestMethod.equalsIgnoreCase("PUT") || originalHttpRequestMethod.equalsIgnoreCase("PATCH")) {
+                                if (httpStatusCode == 201) {
+                                    pollStrategy = AzureAsyncOperationPollStrategy.tryToCreate(AzureProxy.this, fullyQualifiedMethodName, originalHttpResponse, originalHttpRequestUrl, serializer, delayInMilliseconds);
+                                } else if (httpStatusCode == 202) {
+                                    pollStrategy = AzureAsyncOperationPollStrategy.tryToCreate(AzureProxy.this, fullyQualifiedMethodName, originalHttpResponse, originalHttpRequestUrl, serializer, delayInMilliseconds);
+                                    if (pollStrategy == null) {
+                                        pollStrategy = LocationPollStrategy.tryToCreate(AzureProxy.this, fullyQualifiedMethodName, originalHttpResponse, delayInMilliseconds);
+                                    }
+                                }
+                            } else /* if (originalRequestHttpMethod.equalsIgnoreCase("DELETE") || originalRequestHttpMethod.equalsIgnoreCase("POST") */ {
```
RikkiGibson: Is this dead code in the comment? Any ambiguity about whether that condition is needed?

daschult: I originally had it in to make it clear what the else condition was, but I'll take it out.

client-runtime/src/main/java/com/microsoft/rest/http/BufferedHttpResponse.java
```diff
@@ -16,49 +16,72 @@
  * HTTP response which contains already-buffered body content.
```
RikkiGibson: Consider changing this to convey that body content is cached after being read once, not that it's already been buffered by virtue of being contained in this object.

PR #257, opened 2017-10-10
Convert RxNetty to native Netty

jianghaolu:


client-runtime/src/main/java/com/microsoft/rest/http/RxNettyResponse.java
```diff
@@ -7,78 +7,64 @@
...
+
 import java.io.InputStream;
 import java.nio.charset.Charset;
+import java.util.List;
+import java.util.Map.Entry;
 
 /**
  * A HttpResponse that is implemented using RxNetty.
  */
 class RxNettyResponse extends HttpResponse {
```
daschult: Rename to NettyResponse?

client-runtime/src/test/java/com/microsoft/rest/RestProxyWithRxNettyTests.java
```diff
@@ -2,14 +2,21 @@
...
-import com.microsoft.rest.http.RxNettyAdapter;
-import com.microsoft.rest.policy.RequestPolicy;
+import com.microsoft.rest.http.NettyAdapter;
+import com.microsoft.rest.policy.LoggingPolicy;
+import com.microsoft.rest.policy.RequestPolicy.Factory;
 
+import java.util.ArrayList;
 import java.util.Collections;
 
 public class RestProxyWithRxNettyTests extends RestProxyTests {
```
daschult: Rename to RestProxyWithNettyTests

client-runtime/src/test/java/com/microsoft/rest/RestProxyWithRxNettyTests.java
```diff
@@ -2,14 +2,21 @@
...
+    private static NettyAdapter adapter = new NettyAdapter(new ArrayList<Factory>() {
+        {
+            add(new LoggingPolicy.Factory(LogLevel.BODY_AND_HEADERS));
+        }}, Collections.<ChannelHandlerConfig>emptyList());
+
     @Override
     protected HttpClient createHttpClient() {
-        return new RxNettyAdapter(Collections.<RequestPolicy.Factory>emptyList(), Collections.<ChannelHandlerConfig>emptyList());
+        return adapter;
```
daschult: Using a shared fixture can lead to tests that are very difficult to debug. Is there a reason you're opting to not create a new adapter for each test?

jianghaolu: Because the channel pool is shared between all the tests this way. Performance is improved. I should be able to create multiple clients yet still let them share the same channel pool.

RikkiGibson:
Travis says:
```
Running com.microsoft.rest.RestProxyWithNettyTests
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
Is something about the CI environment causing Netty to hang?

PR #258, opened 2017-10-10
Making PagedList to work with Java8 stream apis (by ensuring no Concu…

anuchandy:
…rrentModificationException happening in all valid scenarios)

Fixes 
https://github.com/Azure/azure-sdk-for-java/issues/1942
https://github.com/Azure/azure-sdk-for-java/issues/1877
https://github.com/Azure/azure-sdk-for-java/issues/1641

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
@@ -171,6 +171,7 @@ public abstract class PagedList<E> implements List<E> {
                 } else {
                     loadNextPage();
                 }
+                // Recurse until we load a page with non-zero items.
                 return next();
             } else {
                 try {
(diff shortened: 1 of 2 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
```diff
@@ -139,70 +140,122 @@ protected void setCurrentPage(Page<E> currentPage) {
...
         }
 
         @Override
         public E previous() {
-            return itemsListItr.previous();
+            int i = this.nextIndex - 1;
+            if (i < 0) {
+                throw new NoSuchElementException();
+            } else {
+                if (i >= items.size()) {
```
daschult: Simplify this line by merging it with the previous line (`else` becomes an `else if`).

anuchandy: thanks @daschult , fixed.

PR #264, opened 2017-10-13
Add RestResponse classes

daschult:
We've started to have scenarios where certain requests need to expose the HTTP headers from their responses. This pull request adds the `RestResponse<THeader>` and `RestResponseWithBody<THeader,TBody>` classes and integrates them into RestProxy.


client-runtime/src/main/java/com/microsoft/rest/http/HttpHeaders.java
```diff
@@ -107,6 +109,18 @@ private HttpHeader getHeader(String headerName) {
         return headers.get(headerKey);
     }
 
+    /**
+     * Convert this HttpHeaders collection to JSON.
+     * @return The JSON object that contains this HttpHeaders collection's values.
+     */
+    public JSONObject toJSON() {
```
RikkiGibson: Any specific reasoning behind using org.json instead of `implements JsonSerializable` or `@JsonProperty` from Jackson?

daschult: Done.

PR #267, opened 2017-10-17
Convert from wire types to service method return types

daschult:


client-runtime/src/test/java/com/microsoft/rest/http/MockHttpClientTests.java
```diff
@@ -0,0 +1,58 @@
+package com.microsoft.rest.http;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class MockHttpClientTests {
+    @Test
+    public void sendRequestInternalAsyncToBase64UrlBytes() {
```
RikkiGibson: These tests raise a bit of a red flag to me because they're testing the test harness. Ideally the test harness would be so "dumb" that these tests wouldn't be necessary. It looks like this is validating what byte arrays or lists of byte arrays or what have you get created by the "input generator" code.

Basically, I would consider just pasting these literal/constructed values into a completely independent MockHttpClient scoped to a test class which is about RestProxyTypeConversions, and maybe even renaming the existing RestProxyTests to be RestProxyHttpBinTests or something like that.

As a test maintainer, there's a benefit to having mocks which are scoped even to each individual test-- the implementations then look like "did you give me the one value you were supposed to give me? Then here's the one result you're supposed to get--otherwise, here's an error."

daschult: I used these tests to help me debug an issue, so I figured that I'd leave them since they were helpful once. I'm fine with getting rid of them, though.

RikkiGibson: Either way works.

PR #269, opened 2017-10-17
XML support

RikkiGibson:
This PR adds XML support to the runtime.

Added `enum Encoding` to SerializerAdapter with possible values of JSON and XML. The main reason for doing it this way is it's more ergonomic at the usage sites, which typically calculate the appropriate Encoding value to use based on something like the HTTP response Content-Type. See what you think.

The original "serialize" and "deserialize" methods are still there but are marked deprecated. If this is the design we go with, they should probably be removed soon.

Things to consider:
- RestProxy has a new method: `SerializerAdapter.Encoding bodyEncoding(HttpHeaders)`. I don't know if it's quite the appropriate place. There are a lot of overloads for deserialization in RestProxy itself and I added the Encoding parameter to them.
- The XML tests are a little shaky because I wasn't sure how to test that XML is formed the way I care about. Jackson will happily deserialize XML of many forms into a List, but when we give XML to a service we want to make sure the name of the root element is correct.


client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -393,17 +396,29 @@ public HttpResponse call(String responseBody) {
...
                     });
         }
         return asyncResult;
     }
 
+    private SerializerAdapter.Encoding bodyEncoding(HttpHeaders headers) {
+        String mimeContentType = headers.value("Content-Type");
+        if (mimeContentType != null) {
+            String[] parts = mimeContentType.split(";");
+            if (parts[0].equalsIgnoreCase("application/xml") || parts[0].equalsIgnoreCase("text/xml")) {
```
daschult: Should we be checking all of the parts of the mimeContentType instead of just the first one?

RikkiGibson: If the response content type is really a semicolon-separated list, then yes. However, it intuitively seems to me like it's not valid for there to be multiple content types on a response body, and the use of the semicolon is really to specify stuff like character encodings.

PR #271, opened 2017-10-18
set() Content-Length header instead of add(), handle no response Content-Length header

daschult:


daschult:
This pull request's changes got absorbed into the [AddMapConversions ](https://github.com/Azure/autorest-clientruntime-for-java/pull/272) branch pull request.

PR #272, opened 2017-10-18
Add map conversions

daschult:


client-runtime/src/main/java/com/microsoft/rest/http/NettyResponse.java
```diff
@@ -29,10 +29,20 @@
 
     NettyResponse(io.netty.handler.codec.http.HttpResponse rxnRes, Observable<ByteBuf> emitter) {
         this.rxnRes = rxnRes;
-        this.contentLength = Long.parseLong(rxnRes.headers().get(HEADER_CONTENT_LENGTH));
+        this.contentLength = getContentLength(rxnRes);
```
RikkiGibson: Just curious, was headers() coming up null at any point?

daschult: Not for me.

PR #273, opened 2017-10-19
Consolidate RestResponse

RikkiGibson:
- Now only one RestResponse type which allows any combination of body/headers, only headers, only body, or neither to be specified. Raw headers are always present.

- Minor RestProxy refactor to enforce a notion of "unwrap the Single if it's there, unwrap the RestResponse if it's there, and now handle the body".


client-runtime/src/main/java/com/microsoft/rest/RestProxy.java
```diff
@@ -353,7 +353,34 @@ public HttpResponse call(String responseBody) {
         return asyncResult;
     }
 
-    private Single<?> toProxyReturnValueAsync(final HttpResponse response, final SwaggerMethodParser methodParser, final Type entityType) {
+    private Single<?> handleRestResponseReturnTypeAsync(final HttpResponse response, final SwaggerMethodParser methodParser, final Type entityType) {
```
daschult: methodParser doesn't need to be `final` anymore.

daschult: Neither does response.

daschult: And neither does entityType.

RikkiGibson: Wouldn't it be nice if final were the default?

PR #274, opened 2017-10-20
Remove HttpHeaders.add()

RikkiGibson:
This PR removes the add() method in favor of the set() method because add() has proven to be a behavior that people generally pick by accident and don't really intend. Other minor changes were made in order to fix the autorest.java headers tests.

We should consider making HttpHeaders immutable, as well as HttpRequest itself, as opposed to providing methods for deep copying.


client-runtime/src/main/java/com/microsoft/rest/http/HttpHeaders.java
```diff
@@ -77,7 +59,7 @@ public HttpHeaders add(String headerName, String headerValue) {
      */
     public HttpHeaders set(String headerName, String headerValue) {
         final String headerKey = headerName.toLowerCase();
-        if (headerValue == null || headerValue.isEmpty()) {
```
daschult: Are we allowing empty header values?

PR #277, opened 2017-10-23
Fix LRO tests in the autorest.java repository

daschult:


azure-client-runtime/src/main/java/com/microsoft/azure/AsyncOperationResource.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public String status() {
+        return status;
+    }
+
+    /**
+     * Set the status of the asynchronous operation.
+     * @param status The status of the asynchronous operation.
+     */
+    public void status(String status) {
```
RikkiGibson: `setStatus`?

azure-client-runtime/src/main/java/com/microsoft/azure/AzureAsyncOperationPollStrategy.java
```diff
@@ -73,37 +75,50 @@ else if (pollingSucceeded) {
 
     @Override
     public Single<HttpResponse> updateFromAsync(final HttpResponse httpPollResponse) {
+        if (400 <= httpPollResponse.statusCode()) {
```
RikkiGibson: Not a big thing but I encourage putting constants on the right side of a comparison

azure-client-runtime/src/main/java/com/microsoft/azure/AzureProxy.java
```diff
@@ -140,72 +141,108 @@ protected Object handleAsyncHttpResponse(final HttpRequest httpRequest, Single<H
...
+
+    private Single<PollStrategy> createProvisioningStateOrCompletedPollStrategy(final HttpRequest httpRequest, HttpResponse httpResponse, SwaggerMethodParser methodParser, final long delayInMilliseconds) {
+        Single<PollStrategy> result;
+
+        final String httpRequestMethod = httpRequest.httpMethod();
+        if (httpRequestMethod.equalsIgnoreCase("DELETE") ||
+                httpRequestMethod.equalsIgnoreCase("GET") ||
+                httpRequestMethod.equalsIgnoreCase("HEAD") ||
+                !methodParser.expectsResponseBody()) {
+            result = Single.<PollStrategy>just(new CompletedPollStrategy(AzureProxy.this, httpResponse));
```
RikkiGibson: consider checking content-type: application/json here

RikkiGibson: This way we won't have to buffer octet streams and attempt to parse them as JSON before we give them over

RikkiGibson: My only fear is that autorest.testserver isn't giving a response Content-Type back--but we'll be able to work around it, maybe even with a requestpolicy that adds it in the worst case

azure-client-runtime/src/main/java/com/microsoft/azure/LocationPollStrategy.java
```diff
@@ -64,17 +69,41 @@ public boolean isDone() {
...
-    static PollStrategy tryToCreate(RestProxy restProxy, String fullyQualifiedMethodName, HttpResponse httpResponse, long delayInMilliseconds) {
+    static PollStrategy tryToCreate(RestProxy restProxy, HttpRequest originalHttpRequest, HttpResponse httpResponse, long delayInMilliseconds) {
         final String locationUrl = httpResponse.headerValue(HEADER_NAME);
-        return locationUrl != null && !locationUrl.isEmpty()
-                ? new LocationPollStrategy(restProxy, fullyQualifiedMethodName, locationUrl, delayInMilliseconds)
-                : null;
+
+        String pollUrl = null;
+        if (locationUrl != null && !locationUrl.isEmpty()) {
+            if (locationUrl.startsWith("/")) {
```
RikkiGibson: can consider using https://stackoverflow.com/questions/8273019/how-to-merge-two-uris-in-java#8273086 to combine a base URL with a relative URL

azure-client-runtime/src/main/java/com/microsoft/azure/AsyncOperationResource.java
daschult: Done.

azure-client-runtime/src/main/java/com/microsoft/azure/AzureAsyncOperationPollStrategy.java
daschult: Done.

azure-client-runtime/src/main/java/com/microsoft/azure/LocationPollStrategy.java
daschult: Done.

daschult:
@RikkiGibson Thanks for fixing my checkstyle issues. :)

PR #278, opened 2017-10-27
Stream HTTP response body and accept file HTTP request body

RikkiGibson:
This adds an Observable<byte[]> overload to stream HTTP response content. It also tweaks RestProxy to accept a FileSegment by putting it into a FileRequestBody.


RikkiGibson:
I am still experimenting with the performance/tradeoffs with ByteBuffer vs byte[] as well as how to buffer/backpressure reasonably while we wait for the user to subscribe to the Observable body. Will conclude on it in a future PR.

PR #281, opened 2017-10-30
Migrate to v2 package and add v2 to groupId in POMs

RikkiGibson:
We've determined that renaming the package for v2 is necessary for multiple reasons.

1. Necessary to use SDKs depending on runtime v1 and runtime v2 side-by-side
2. Allows us to publish 2.0.0-SNAPSHOT packages for external teams without breaking builds for everyone specifying a [1.1.0, 2.0.0) version range for the runtime in their pom.

This PR migrates all the source files to the new package path and modifies the groupID in all the POMs.


azure-client-authentication/pom.xml
```diff
@@ -6,7 +6,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>com.microsoft.azure</groupId>
+    <groupId>com.microsoft.azure.v2</groupId>
```
daschult: Let's get Martin and Asir's opinion on this before merging it in. I just want to make sure that they're on board with this.

RikkiGibson:
@martinsawicki @selvasingh Any concerns about the use of `com.microsoft.azure.v2` and `com.microsoft.rest.v2` as the new package paths?

PR #282, opened 2017-10-30
Make HttpRequest mutable

RikkiGibson:


client-runtime/src/test/java/com/microsoft/rest/v2/http/HttpRequestTests.java
```diff
@@ -12,4 +15,40 @@ public void constructor() {
...
+            for (HttpHeader originalHeader : request.headers()) {
+                if (clonedHeader == originalHeader) {
+                    fail("Failed to clone HttpHeader instance");
+                }
+            }
+
+            assertEquals(clonedHeader.value(), request.headers().value(clonedHeader.name()));
+        }
+
+        assertEquals(request.body(), clonedRequest.body());
```
daschult: You might want to add some assertions about same-ness (`assertSame()` or `assertNotSame()`) in this test too.

RikkiGibson: Looked up assertSame and it looks like it indicates more clearly what is under test (the identity of the objects, not the value)

PR #283, opened 2017-10-30
Fix broken links and HTML in Javadocs

RikkiGibson:


RikkiGibson:
Would like to make some of the formatting better in the Javadocs but not a priority right now

PR #284, opened 2017-10-30
Add better error message when a polling strategy cannot be determined

daschult:


RikkiGibson:
Would like to see what this shows when that transient failure occurs again.

PR #290, opened 2017-10-31
Rename RestClient methods

RikkiGibson:
- Renames `addCustomPolicy` to `addRequestPolicy`
- Adds `setRequestPolicies(List<RequestPolicy.Factory)`
- Renames `withCredentials(ServiceClientCredentials)` to `withCredentialsPolicy(RequestPolicy.Factory)`
- Removes `withMaxIdleConnections(int)`

The point of "add" instead of "with" is to convey that repeatedly calling it doesn't cause the previous value to get overwritten. Not sure whether it's more consistent to use the name `withRequestPolicies` instead of `setRequestPolicies`.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/policy/CredentialsPolicy.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/policy/CredentialsPolicy.java
@@ -14,7 +14,7 @@ import rx.Single;
 import java.io.IOException;
 
 /**
- * Adds credentialsPolicyFactory from ServiceClientCredentials to a request.
+ * Adds credentials from ServiceClientCredentials to a request.
  */
 public final class CredentialsPolicy implements RequestPolicy {
     /**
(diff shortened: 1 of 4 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/policy/CredentialsPolicy.java
```diff
@@ -14,7 +14,7 @@
 import java.io.IOException;
 
 /**
- * Adds credentials from ServiceClientCredentials to a request.
+ * Adds credentialsPolicyFactory from ServiceClientCredentials to a request.
```
daschult: I think this might not be the refactoring you were looking for.

client-runtime/src/main/java/com/microsoft/rest/v2/policy/CredentialsPolicy.java
```diff
@@ -25,7 +25,7 @@
 
         /**
          * Creates a Factory which produces CredentialsPolicy.
-         * @param credentials The credentials to use for authentication.
+         * @param credentials The credentialsPolicyFactory to use for authentication.
```
daschult: Bad IntelliJ. Bad!

RikkiGibson: Gotta look at those refactor previews and diffs before just doing it..

PR #291, opened 2017-11-01
Check status code before and during polling strategy

daschult:


azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -141,84 +139,83 @@ protected Object handleAsyncHttpResponse(final HttpRequest httpRequest, Single<H
...
-                            }
-                        }
-                        else {
-                            if (httpStatusCode == 202) {
-                                pollStrategy = AzureAsyncOperationPollStrategy.tryToCreate(AzureProxy.this, originalHttpRequest, originalHttpResponse, expectsResourceResponse, serializer, delayInMilliseconds);
-                                if (pollStrategy == null) {
-                                    pollStrategy = LocationPollStrategy.tryToCreate(AzureProxy.this, originalHttpRequest, originalHttpResponse, delayInMilliseconds);
-                                    if (pollStrategy == null) {
-                                        throw new CloudException("Response does not contain an Azure-AsyncOperation or Location header.", originalHttpResponse);
+                        final int[] longRunningOperationStatusCodes = new int[] {200, 201, 202};
```
RikkiGibson: This is good-- now manually adding extra status codes is no longer necessary or desired in the generator.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -141,84 +139,83 @@ protected Object handleAsyncHttpResponse(final HttpRequest httpRequest, Single<H
...
+                                    public Single<? extends PollStrategy> call(HttpResponse response) {
+                                        Single<PollStrategy> result = null;
+
+                                        final Long parsedDelayInMilliseconds = PollStrategy.delayInMillisecondsFrom(originalHttpResponse);
+                                        final long delayInMilliseconds = parsedDelayInMilliseconds != null ? parsedDelayInMilliseconds : AzureProxy.defaultDelayInMilliseconds();
+
+                                        final String originalHttpRequestMethod = originalHttpRequest.httpMethod();
+
+                                        PollStrategy pollStrategy = null;
+                                        if (httpStatusCode == 200) {
```
RikkiGibson: It's probably not desirable to do the work to change it now, but I think a more repetitive style of conditions in if statements could be easier to read. Each if statement you go deeper into requires you to accumulate in your mind what assumptions can be made inside the block, and inside that child block, etc.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/LocationPollStrategy.java
```diff
@@ -29,34 +30,39 @@
...
-            String newLocationUrl = httpPollResponse.headerValue(HEADER_NAME);
-            if (newLocationUrl != null) {
-                locationUrl = newLocationUrl;
-            }
-        }
-        else {
-            done = true;
-        }
-        return Single.just(httpPollResponse);
+        return ensureExpectedStatus(httpPollResponse, new int[] {202})
```
RikkiGibson: Small, maybe immeasurable thing: these kinds of constant arrays can be static final to reduce allocations.

PR #293, opened 2017-11-02
[Do not merge] Proxy

RikkiGibson:
Opening this early to see what the proxy tests do in CI.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -6,7 +6,6 @@
 
 package com.microsoft.rest.v2.http;
 
-import com.microsoft.rest.v2.policy.RequestPolicy;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
@@ -33,9 +32,11 @@ import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.subjects.ReplaySubject;
 
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.net.SocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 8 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -98,16 +100,34 @@ public void channelCreated(Channel ch) throws Exception {
...
+                } else if (proxy.address() instanceof InetSocketAddress) {
+                    InetSocketAddress address = (InetSocketAddress) proxy.address();
+                    String scheme = address.getPort() == 443
+                            ? "https"
+                            : "http";
+
+                    String urlString = scheme + "://"
+                            + ((InetSocketAddress) proxy.address()).getHostString()
+                            + ":" + address.getPort();
+                    uri = new URI(urlString);
```
daschult: This currently drops any path and/or query parameters that exist in the request's url. I'd recommend using our UrlBuilder to create the url. That should make it easier to construct the proxy url.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -98,16 +100,34 @@ public void channelCreated(Channel ch) throws Exception {
...
+                if (proxy == null) {
+                    uri = new URI(request.url());
+                } else if (proxy.address() instanceof InetSocketAddress) {
+                    InetSocketAddress address = (InetSocketAddress) proxy.address();
+                    String scheme = address.getPort() == 443
+                            ? "https"
+                            : "http";
+
+                    String urlString = scheme + "://"
+                            + ((InetSocketAddress) proxy.address()).getHostString()
```
daschult: Reuse `address` from line 109 instead of casting again.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -98,16 +100,34 @@ public void channelCreated(Channel ch) throws Exception {
...
+                            + ":" + address.getPort();
+                    uri = new URI(urlString);
+                } else {
+                    throw new IllegalArgumentException(
+                            "SocketAddress on java.net.Proxy must be an InetSocketAddress. Instead found "
+                                    + proxy.address());
```
daschult: You may want to write out the address's type in this exception too.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: This uri is just used to acquire a channel, so it only needs the information needed to establish a TCP connection. Do you still want me to append the path and query? I actually considered refactoring the pool to make it use SocketAddress instead of URI, but I didn't want to change any host resolution behaviors @jianghaolu may have intended.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: Proxy.toString() might be suitable here--implementation appears to show both the type and address.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: Good catch.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: For now I'm renaming the variable to "channelAddress" for clarify-- we may wish to change the pool to acquire based on SocketAddress in a separate PR.

PR #294, opened 2017-11-03
Migrate to RxJava2

RikkiGibson:
This grinds out all the package and API changes related to RxJava 2. We'd like to be able to use the backpressure features of Flowable and customers have asked us for RxJava 2 support as well. Similar work will need to happen in autorest.java and in any dependent projects.

Actually changing some of these Observables to Flowables, e.g. HttpResponse.streamBodyAsync, will happen a little bit later.


RikkiGibson:
Any refactoring work in autorest.java is probably higher priority than fixing that to work with this.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
```diff
@@ -138,20 +140,20 @@ void setStatus(String status) {
...
-        return Observable.defer(new Func0<Observable<HttpResponse>>() {
+        return Observable.defer(new Callable<Observable<HttpResponse>>() {
             @Override
             public Observable<HttpResponse> call() {
                 return delayAsync()
-                        .flatMap(new Func1<Void, Single<HttpResponse>>() {
+                        .andThen(Single.defer(new Callable<Single<HttpResponse>>() {
                             @Override
-                            public Single<HttpResponse> call(Void ignored) {
+                            public Single<HttpResponse> call() throws Exception {
```
daschult: Is the `throws Exception` part of the `Callable`'s `call()` method signature?

azure-client-runtime/src/test/java/com/microsoft/azure/v2/AzureProxyToRestProxyTests.java
```diff
@@ -447,47 +447,24 @@ public void SyncPutRequestWithUnexpectedResponseAndExceptionType() {
...
-        @ExpectedResponses({200})
         Completable completableHeadAsync();
     }
 
     @Test
-    public void SyncHeadRequest() {
-        final HttpBinJSON json = createService(Service10.class)
-                .head();
-        assertNull(json);
-    }
```
daschult: Why are we getting rid of some of these tests?

azure-client-runtime/src/test/java/com/microsoft/azure/v2/PagedListTests.java
```diff
@@ -279,30 +279,30 @@ public void canCreateObservableFromPagedList() {
                 return firstObservable().concatWith(nextObservable());
             }
 
-            rx.Observable<Integer> firstObservable() {
-                return rx.Observable.defer(new Func0<rx.Observable<Integer>>() {
+            io.reactivex.Observable<Integer> firstObservable() {
```
daschult: Why do we need to include the package here?

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -317,7 +317,7 @@ private Exception instantiateUnexpectedException(SwaggerMethodParser methodParse
             final Constructor<? extends RestException> exceptionConstructor = exceptionType.getConstructor(String.class, HttpResponse.class, exceptionBodyType);
             final Object exceptionBody = responseContent == null || responseContent.isEmpty() ? null : serializer.deserialize(responseContent, exceptionBodyType, bodyEncoding(response.headers()));
 
-            result = exceptionConstructor.newInstance("Status code " + responseStatusCode + ", " + responseContent, response, exceptionBody);
+            result = exceptionConstructor.newInstance("Status code " + responseStatusCode + ", " + (exceptionBody == null ? null : responseContent), response, exceptionBody);
```
daschult: This change is essentially saying that if `exceptionBody` is null, then we're going to throw an exception that says that the `responseContent` is null. However, that's not entirely true, because it's possible that the deserializer returned null when the `responseContent` is not null and not empty. What was your intention with this change?

client-runtime/src/test/java/com/microsoft/rest/v2/RestProxyTests.java
```diff
@@ -528,7 +526,7 @@ public void AsyncPutRequestWithUnexpectedResponseAndExceptionType() {
     private interface Service10 {
         @HEAD("anything")
         @ExpectedResponses({200})
-        HttpBinJSON head();
+        RestResponse<Void, Void> head();
```
daschult: Why did you decide to change these tests?

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
RikkiGibson: Yeah, it looks like when you use IntelliJ's "implement methods" feature, it adds it to the signature.

azure-client-runtime/src/test/java/com/microsoft/azure/v2/AzureProxyToRestProxyTests.java
RikkiGibson: `Observable<Void>` and `Single<Void>` are now illegal, so I'll bring this back as `Single<RestResponse<Void, Void>>`. I think I did that in the other tests, but not here..

azure-client-runtime/src/test/java/com/microsoft/azure/v2/PagedListTests.java
RikkiGibson: We don't. Seems like it ended up there by mistake. (is there checkstyle for qualified access when the class has been imported already?)

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
RikkiGibson: This is dealing with the fact that you can't emit a null from `HttpResponse.getBodyAsStringAsync()` or any of the other variants. Actually, the more I think about this code, the more I realize it needs to be refactored. I think this might be related to why I'm getting really big, wacky console output when a file download fails in testing.

RikkiGibson:
Waiting to merge until we've determined that this is going to work for our users.

client-runtime/src/test/java/com/microsoft/rest/v2/RestProxyTests.java
RikkiGibson: Can't emit `null` from a stream any more in RxJava 2, so in general HEAD requests that return a body entity type aren't really valid any more.

PR #297, opened 2017-11-07
Migrate to RxJava2

RikkiGibson:
Continuation of #294, but based on Azure:RxJava2 instead of rikkigibson:RxJava2 to make it easier for others to work on it.

> This grinds out all the package and API changes related to RxJava 2. We'd like to be able to use the backpressure features of Flowable and customers have asked us for RxJava 2 support as well. Similar work will need to happen in autorest.java and in any dependent projects.
> 
> Actually changing some of these Observables to Flowables, e.g. HttpResponse.streamBodyAsync, will happen a little bit later.


RikkiGibson:
Also note that there is code generator work that needs to be completed before any snapshots should be published using this PR. See https://github.com/azure/autorest.java/tree/RxJava2. Mainly, the difficulty is in determining how to stamp out any uses of `Single<Void>`.

RikkiGibson:
This is mostly ready to go--the only questionable part is the fact that if a response body might deserialize to null you need to use the `WithRestResponseAsync` generated method so that a null doesn't get emitted in the stream.

I'm considering using `Maybe<Entity> getAsync` so those cases can be handled better, but I want to be a little more confident in the decision before I start changing the code generator to produce that.

RikkiGibson:
Maybe<Entity> is the new signature.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -313,17 +313,34 @@ private Exception instantiateUnexpectedException(SwaggerMethodParser methodParse
...
             final Constructor<? extends RestException> exceptionConstructor = exceptionType.getConstructor(String.class, HttpResponse.class, exceptionBodyType);
-            final Object exceptionBody = responseContent == null || responseContent.isEmpty() ? null : serializer.deserialize(responseContent, exceptionBodyType, bodyEncoding(response.headers()));
 
-            result = exceptionConstructor.newInstance("Status code " + responseStatusCode + ", " + responseContent, response, exceptionBody);
+            String contentType = response.headerValue("Content-Type");
+            boolean isSerializableContentType = contentType == null || contentType.isEmpty()
+                    || "application/json".equalsIgnoreCase(contentType)
+                    || "text/json".equalsIgnoreCase(contentType)
+                    || "application/xml".equalsIgnoreCase(contentType)
+                    || "text/xml".equalsIgnoreCase(contentType);
```
daschult: Should these be `startsWith()` instead of `equals()` because there could be a `charset` value?

RikkiGibson: Good question--probably should add coverage for such a contentType value, and add coverage for some non-serialized content type like text/html or something.

RikkiGibson:
Added some coverage for exception serialization. Do we want to throw a RuntimeException containing IOException if we fail to create RestException due to a JSON/XML response body failing to parse?

RikkiGibson:
Will continue work on Flowable body content in a future PR as that will take some additional thought and care to get right.

PR #298, opened 2017-11-09
Add delegated token credentials

jianghaolu:


azure-client-authentication/src/main/java/com/microsoft/azure/credentials/ApplicationTokenCredentials.java
```diff
@@ -162,7 +174,7 @@ private AuthenticationResult acquireAccessToken(String resource) throws IOExcept
         }
     }
 
-    private PrivateKey privateKeyFromPem(String pem) {
+    static PrivateKey privateKeyFromPem(String pem) {
```
daschult: What does Pem stand for?

jianghaolu: > PEM or Privacy Enhanced Mail is a Base64 encoded DER certificate.
https://en.wikipedia.org/wiki/Privacy-enhanced_Electronic_Mail

PR #302, opened 2017-11-20
Replace RestClient with HttpPipeline

daschult:
After the recent discussion that Jianghao and I had with Jeffrey and Josh, we determined that there are some changes to our runtime that needed to be made with regard to RestClient and the RequestPolicy pipeline. This pull request implements those changes.

First of all, this pull request introduces the HttpPipeline object. This object is an immutable list of RequestPolicy.Factory objects with a HttpClient at the end. It also contains an asynchronous sendRequest(HttpRequest) method that will instantiate and chain each of the RequestPolicies together, and then it will send the provided HttpRequest to the chain. I've extracted the RequestPolicy.Factory list from HttpClient class and put it here instead. In this way, HttpPipeline wraps around an HttpClient object.

Second, this pull request removes RestClient and AzureRestClient. AzureRestClient was really just a pass-through to RestClient, so it could easily be removed. RestClient was a grab-bag of configuration values, but those values can easily be moved to RequestPolicys within the pipeline.


client-runtime/src/main/java/com/microsoft/rest/v2/annotations/HostParam.java
```diff
@@ -52,5 +52,5 @@
      * encoded hence engine should not encode it, by default value will be encoded.
      * @return Whether or not this argument is already encoded.
      */
-    boolean encoded() default false;
+    boolean encoded() default true;
```
jianghaolu: Why change to true?

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpPipeline.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Build a new HttpPipeline that will use the provided HttpClient and RequestPolicy factories.
+     * @param requestPolicyFactory1 The first RequestPolicy factory to use.
+     * @param requestPolicyFactory2 The second RequestPolicy factory to use.
+     * @param httpClient The HttpClient to use.
+     * @return The built HttpPipeline.
+     */
+    public static HttpPipeline build(RequestPolicy.Factory requestPolicyFactory1, RequestPolicy.Factory requestPolicyFactory2, HttpClient httpClient) {
```
jianghaolu: You can use `build(HttpClient httpClient, RequestPolicy.Factory... factories)`. It accepts an array too.

client-runtime/src/main/java/com/microsoft/rest/v2/annotations/HostParam.java
daschult: Because if this was set to false, then if my interface had a Host annotation like `@Host("{baseUrl}")` and a function with a HostParam annotation like `@HostParam("baseUrl")` where the HostParam's value was "https://my.base.url", then the "://" would get encoded. That was breaking some of our internal logic. I think it's more likely that HostParam values will be passed in already encoded.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpPipeline.java
daschult: I changed the order of the parameters for that function. I felt that it better reflected the flow that HttpRequests would take as they go through the HttpPipeline. They start with the RequestPolicies and then end up at the HttpClient. That's the reason I added those overloads that take a fixed number of RequestPolicy.Factory objects. It's an attempt to try to make it easier to use and to visualize the pipeline.

PR #303, opened 2017-11-21
Add UrlBuilder.parse()

daschult:
I added this method because I need to modify a URL in autorest.java and I need to check if a port already exists. If it doesn't, then I need to add it. java.net.URL doesn't allow me to check if the port already exists, otherwise I would've just used that.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/policy/PortPolicy.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/policy/PortPolicy.java
@@ -30,7 +30,7 @@ public class PortPolicy implements RequestPolicy {
     public Single<HttpResponse> sendAsync(HttpRequest request) {
         final UrlBuilder urlBuilder = UrlBuilder.parse(request.url());
         if (overwrite || urlBuilder.port() == null) {
-            request.withUrl(urlBuilder.withPort(3000).toString());
+            request.withUrl(urlBuilder.withPort(port).toString());
         }
         return nextPolicy.sendAsync(request);
     }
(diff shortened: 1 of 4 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/policy/PortPolicy.java
```diff
@@ -0,0 +1,68 @@
...
+        this.nextPolicy = nextPolicy;
+        this.port = port;
+        this.overwrite = overwrite;
+    }
+
+    @Override
+    public Single<HttpResponse> sendAsync(HttpRequest request) {
+        final UrlBuilder urlBuilder = UrlBuilder.parse(request.url());
+        if (overwrite || urlBuilder.port() == null) {
+            request.withUrl(urlBuilder.withPort(3000).toString());
```
jianghaolu: default to 3000?

daschult: Copy and paste mistake. Fixed.

RikkiGibson:
It sounds like this would be useful for cases where we have to merge additional parameters into an existing URL. But it also looks like java.net.URL.getPort() returns -1 when the port is not specified in syntax in the URL.

```
    @Test
    public void urlTest() throws Exception {
        java.net.URL url = new java.net.URL("http://microsoft.com");
        assertEquals(80, url.getPort());
    }

    // AssertionError: Expected 80, Actual -1
```

daschult:
@RikkiGibson Yeah, it would be useful for those scenarios. I originally attempted to write my own URL parser, but it just got to be more work than it was worth. I discovered what you mentioned about java.net.URL returning -1 if a port wasn't specified, so I ended up just wrapping it's constructor/parse() method.

PR #307, opened 2017-11-28
Add extra request policies to RestProxy's default pipeline

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpClient.java
```diff
@@ -21,12 +21,8 @@
      */
     public abstract Single<HttpResponse> sendRequestAsync(HttpRequest request);
 
-    private static HttpClient.Factory defaultFactory;
     private static HttpClient.Factory defaultHttpClientFactory() {
-        if (defaultFactory == null) {
-            defaultFactory = new NettyClient.Factory();
-        }
-        return defaultFactory;
+        return new NettyClient.Factory();
```
daschult: @jianghaolu I ran into issues in the autorest.java vanilla tests when I was using a shared factory/adapter. A few different tests would just hang, specifically after a multi-threaded test had been run. When I removed the shared factory/adapter, then the tests started passing. Just thought you'd like to be aware of this.

PR #309, opened 2017-11-29
Create HttpPipeline.Builder from an existing HttpPipeline

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpPipeline.java
```diff
@@ -167,7 +222,90 @@ public Builder withLogger(Logger logger) {
...
+                    ++searchIndex;
+                }
+            }
+            final int factoryCount = requestPolicyFactories.size();
+
+            if (searchIndex == factoryCount) {
+                withRequestPolicy(requestPolicyFactory);
+            } else {
+                final int insertIndex = factoryCount - searchIndex;
+                withRequestPolicy(insertIndex, requestPolicyFactory);
```
RikkiGibson: This expands to an insert at (factoryCount - (factoryCount - searchIndex)) which seems a little obtuse. Just insert into requestPolicyFactories.

RikkiGibson: It's a tricky thing to keep straight in general, though, because it's natural to want to keep the factories in "sequential" order from an external POV while reversing them at the appropriate time to build the linked list, etc.

daschult: Done.

PR #310, opened 2017-11-30
Add AzureEnvironment to AzureServiceClient and add AzureProxy.create(…

daschult:
…) overload that takes an AzureServiceClient

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
@@ -203,7 +203,7 @@ public final class AzureProxy extends RestProxy {
      */
     @SuppressWarnings("unchecked")
     public static <A> A create(Class<A> swaggerInterface, HttpPipeline httpPipeline, SerializerAdapter<?> serializer) {
-        return AzureProxy.create(swaggerInterface, (AzureEnvironment) null, httpPipeline, serializer);
+        return AzureProxy.create(swaggerInterface, null, httpPipeline, serializer);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -182,39 +182,51 @@ public static HttpPipeline defaultPipeline(Class<?> swaggerInterface, RequestPol
...
+     * @param azureEnvironment The azure environment that the proxy implementation will target.
+     * @param httpPipeline The HTTP httpPipeline will be used to make REST calls.
      * @param serializer The serializer that will be used to convert POJOs to and from request and
      *                   response bodies.
      * @param <A> The type of the Swagger interface.
      * @return A proxy implementation of the provided Swagger interface.
      */
     @SuppressWarnings("unchecked")
-    public static <A> A create(Class<A> swaggerInterface, String baseUrl, HttpPipeline httpPipeline, SerializerAdapter<?> serializer) {
+    public static <A> A create(Class<A> swaggerInterface, AzureEnvironment azureEnvironment, HttpPipeline httpPipeline, SerializerAdapter<?> serializer) {
```
RikkiGibson: Are you changing the generator to pass AzureEnvironment down so that RestProxy can do the lookup here? Also, was it basically decided that providing the base URL directly is unnecessary for any Azure use case?

daschult: Providing a baseUrl is unnecessary now. We have both the Host and AzureHost annotations which provide all of the functionality that baseUrl used to provide.
RestProxy never needs to know about baseUrl. Only SwaggerInterfaceParser needs the baseUrl, and it gets that from either AzureProxy (parsing the AzureHost annotation) or from its own parsing of the Host annotation.

PR #311, opened 2017-12-01
Add HostPolicy

daschult:


RikkiGibson:
Could you add test coverage for a host with a port specified? Note that the value for a `@Host` annotation comes from the Swagger `host` attribute, which can contain a port number. https://swagger.io/docs/specification/2-0/api-host-and-base-path/

daschult:
@RikkiGibson Good request. Done.

PR #313, opened 2017-12-05
Add http scheme if no scheme is provided when creating a HttpRequest …

daschult:
…in RestProxy

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
@@ -250,6 +250,11 @@ public class RestProxy implements InvocationHandler {
             urlBuilder.withQueryParameter(queryParameter.name(), queryParameter.encodedValue());
         }
 
+        final String scheme = urlBuilder.scheme();
+        if (scheme == null || scheme.isEmpty()) {
+            urlBuilder.withScheme("http");
+        }
+
         final String url = urlBuilder.toString();
         final HttpRequest request = new HttpRequest(methodParser.fullyQualifiedMethodName(), methodParser.httpMethod(), url);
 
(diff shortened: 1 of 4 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -32,6 +32,7 @@
 import rx.Single;
 import rx.exceptions.Exceptions;
 import rx.functions.Func1;
+import sun.plugin.dom.exception.InvalidStateException;
```
RikkiGibson: Checkstyle might be mad about this?

daschult: I always forget to run checkstyle after I make little changes to a pull request. I'll fix this.

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodParser.java
```diff
@@ -220,7 +220,7 @@ public String httpMethod() {
     public String scheme(Object[] swaggerMethodArguments) {
         final String substitutedHost = applySubstitutions(rawHost, hostSubstitutions, swaggerMethodArguments, UrlEscapers.urlPathSegmentEscaper());
         final String[] substitutedHostParts = substitutedHost.split("://");
-        return substitutedHostParts.length < 1 ? "https" : substitutedHostParts[0];
+        return substitutedHostParts.length < 1 ? null : substitutedHostParts[0];
```
RikkiGibson: Is the plan now that we're not going to pick a default scheme for users? As in, if Swagger didn't provide it and the host parameter value didn't provide it, then we just pass it on to the HTTP client and it does whatever it decides to do?

daschult: Yes. We have some tests in autorest.java that will run into this very scenario, which is why I created the ProtocolPolicy RequestPolicy. If no protocol is provided and the HttpRequest hits the HttpClient, then the HttpClient will most likely throw an exception.

RikkiGibson:
Sounds good.

PR #314, opened 2017-12-05
Parse query out of host

RikkiGibson:


RikkiGibson:
@daschult is generalizing this

PR #319, opened 2017-12-08
Recent requests from Rikki and Josh

daschult:


client-runtime/src/test/java/com/microsoft/rest/v2/RestProxyTests.java
```diff
@@ -1010,14 +1009,14 @@ public void service19PutWithHeaderApplicationJsonContentTypeAndCharsetAndStringB
...
-        assertEquals("\"\"", result.data);
+        assertEquals("", result.data);
     }
 
     @Test
     public void service19PutWithHeaderApplicationJsonContentTypeAndCharsetAndStringBodyWithNonEmptyBody() {
         final HttpBinJSON result = createService(Service19.class)
                 .putWithHeaderApplicationJsonContentTypeAndCharsetAndStringBody("soups and stuff");
-        assertEquals("\"soups and stuff\"", result.data);
+        assertEquals("soups and stuff", result.data);
```
RikkiGibson: This is because RestProxy is no longer serializing the String before putting it in the request body?

RikkiGibson:
Just need to merge the Flowable work and fix build errors in there (apparently I left a no-argument BodyParam in there?)

PR #320, opened 2017-12-11
Hide HttpPipeline.Logger inside RequestPolicy.Options

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/policy/RequestPolicy.java
```diff
@@ -58,11 +58,34 @@ public Options(HttpPipeline.Logger logger) {
...
          */
-        public HttpPipeline.Logger logger() {
-            return logger;
+        public boolean shouldLog(HttpPipeline.LogLevel logLevel) {
+            boolean result = false;
+
+            if (logger != null && logLevel != null && logLevel != HttpPipeline.LogLevel.OFF) {
+                final HttpPipeline.LogLevel minimumLogLevel = logger.minimumLogLevel();
+                if (minimumLogLevel != null) {
+                    result = logLevel.ordinal() <= minimumLogLevel.ordinal();
```
RikkiGibson: Is there test coverage on this method?

daschult: Thanks for asking. There is now.

PR #321, opened 2017-12-11
Extract all inner classes to their own files

daschult:
Josh and I discussed this and decided that the inner classes were getting annoying/frustrating. We both agreed that it'd be better to extract them into their own files.


RikkiGibson:
I think because of the IDE behaviors around it this makes more sense. It might also be appropriate for us to start having RequestPolicyFactory implementations as public classes with the corresponding RequestPolicy implementation as a private inner class.

RikkiGibson:
All that concerns me is the churn-- do you mind doing the PR that fixes the breaking changes resulting from this in autorest.java?

daschult:
I've added the fix for the generator [here](https://github.com/Azure/autorest.java/pull/57). I really didn't think there'd need to be any changes, but I forgot about the tests that still use the Builder.

PR #322, opened 2017-12-12
Use Flowable HTTP request body

RikkiGibson:
This PR removes usage of InputStream from HTTP request bodies, replacing it with Flowable<byte[]>.

It appears that making use of multiple implementations of HttpRequestBody requires we generate additional overloads with corresponding parameter types, where RestProxy converts that parameter type to a particular HttpRequestBody type. Since we want to keep additional overloads to a minimum except when a significant performance improvement is possible, I've removed:

- ByteArrayRequestBody which corresponded to byte[]
- FileRequestBody which corresponded to FileSegment, and
- InputStreamRequestBody which corresponded to some combination of InputStream and a length value.

RestProxy still supports byte[] and InputStream in proxy interfaces but it simply does the conversion to FlowableHttpRequestBody one way or another for you.

Now, we have a new "corresponding" type which will appear in generated interfaces called AsyncInputStream, which is a simple pair of a Flowable<byte[]> and a length value. There are convenience `create` methods to convert a file or an InputStream into AsyncInputStream, so autorest.java users will be able to bring all their existing data sources to the table that they had before this change.

I'd like to keep the door open for a second HttpRequestBody implementation which uses zero-copy I/O, and perhaps generation of additional overloads to match. But beyond that it seems like it's not worth the complexity to add more than that. It's conceivable that this could even be done in NettyClient just by looking at the polymorphic type of the content Flowable (are you a FlowableFileStream? OK, I'll create a Netty FileRegion and try to do zero-copy) but it seems somehow unclean to do it that way.

Remaining:
- [x] LoggingPolicy needs to be fixed because of the changes in streaming request content. Also want to take an opportunity to add improvements, like not logging octet-stream content and not logging content over a particular size.

- [x] NettyClient should not request `Long.MAX_VALUE` from the request content Flowable. I think this risks overwhelming Netty as the source Flowable buffers large numbers of events. It might be necessary to start using an OutboundChannelHandler to adapt the request content Flowable and handle `channelWritabilityChanged`. Would like to examine how this might affect upload performance.


client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -306,11 +307,15 @@ private HttpRequest createHttpRequest(SwaggerMethodParser methodParser, Object[]
...
+                AsyncInputStream stream = (AsyncInputStream) bodyContentObject;
+                request.withBody(new FlowableHttpRequestBody(stream.contentLength(), contentType, stream.content()));
+            }
             else if (bodyContentObject instanceof byte[]) {
                 request.withBody((byte[]) bodyContentObject, contentType);
             }
             else if (bodyContentObject instanceof FileSegment) {
-                request.withBody(new FileRequestBody((FileSegment) bodyContentObject));
+                throw new UnsupportedOperationException("FileSegment/zero-copy I/O is not currently supported.");
             }
```
daschult: If we're not going to support it right now, then I'd suggest completely removing the FileSegment class. We can always look in the history if we need to look up how the implementation worked. It might be worth adding a tag for this commit in git that specifies that the implementation of FileSegment can be found here.

client-runtime/src/main/java/com/microsoft/rest/v2/http/AsyncInputStream.java
```diff
@@ -0,0 +1,118 @@
...
+                            }
+                        } catch (IOException e) {
+                            emitter.onError(e);
+                        }
+                    }
+                },
+                new Consumer<InputStream>() {
+                    @Override
+                    public void accept(InputStream inputStream) throws Exception {
+                        inputStream.close();
```
daschult: Should we be closing this InputStream since it is provided by the user?

client-runtime/src/main/java/com/microsoft/rest/v2/http/FlowableFileStream.java
```diff
@@ -0,0 +1,99 @@
...
+            public void request(long n) {
+                chunksRequested += n;
+                readSegmentsAsync();
+            }
+
+            void readSegmentsAsync() {
+                fileChannel.read(innerBuf, position, null, onReadComplete);
+            }
+
+            final CompletionHandler<Integer, Object> onReadComplete = new CompletionHandler<Integer, Object>() {
```
daschult: Any reason this is package private and not private?

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
RikkiGibson: That sounds like the right thing to do.

client-runtime/src/main/java/com/microsoft/rest/v2/http/FlowableFileStream.java
RikkiGibson: `private` helps to convey intent, but non-overriding members of anonymous classes aren't accessible from outside. Might as well have it, though.

client-runtime/src/main/java/com/microsoft/rest/v2/http/AsyncInputStream.java
RikkiGibson: Yeah, in general it seems like we should keep users in control of the lifecycle of resources they provide.

client-runtime/src/main/java/com/microsoft/rest/v2/http/FlowableHttpRequestBody.java
```diff
@@ -42,13 +43,16 @@ public String contentType() {
...
+    public Flowable<byte[]> content() {
         return content;
     }
 
     @Override
     public HttpRequestBody buffer() throws IOException {
-        final byte[] bytes = ByteStreams.toByteArray(content);
-        return new ByteArrayHttpRequestBody(bytes, contentType);
+        if (!isBuffered) {
+            content = content.replay().autoConnect();
```
RikkiGibson: Would be good to have automated test coverage for this. Logging is using it and it's working there but you have to specify `LogLevel.BODY`.

RikkiGibson: Modified a RestProxy test to add this RequestPolicy in such a way that `.replay` is required for the test to complete successfully.

RikkiGibson:
There's still a stability issue when using SSL and getting throttled, but the heart of this is ready now.

PR #331, opened 2017-12-22
Cancellation for file uploads

RikkiGibson:
This makes it so when a Single<HttpResponse> coming from NettyClient is disposed before completion, the channel will be closed, causing the ongoing upload to be canceled.

Also adds several checks to responseEmitter.isDisposed before emitting errors. This is because RxJava will dump big stack traces into your logs if you try to emit an error from a stream that's already been disposed. These are errors people probably won't care about considering they canceled the request that caused the error anyway.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -152,6 +152,7 @@ public final class NettyClient extends HttpClient {
 
                             if (responseEmitter.isDisposed()) {
                                 channel.close();
+                                channelPool.release(channel);
                                 return;
                             }
 
@@ -164,6 +165,7 @@ public final class NettyClient extends HttpClient {
                                     isDisposed = true;
                                     if (!inboundHandler.didEmitHttpResponse) {
                                         channel.close();
+                                        channelPool.release(channel);
                                     }
                                 }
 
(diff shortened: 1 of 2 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -138,13 +140,41 @@ public void subscribe(final SingleEmitter<HttpResponse> responseEmitter) {
                         @Override
                         public void operationComplete(Future<? super Channel> cf) {
                             if (!cf.isSuccess()) {
-                                responseEmitter.onError(cf.cause());
+                                if (!responseEmitter.isDisposed()) {
```
daschult: Having these checks everywhere makes me wonder if we should wrap responseEmitter with an object that will check `isDisposed()` before it makes any call. This just seems like it will be error prone over time.

daschult: Just something to think about.

RikkiGibson: Yeah, that's a good idea.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -371,6 +363,7 @@ void onError(Throwable cause) {
         private ResponseContentFlowable contentEmitter;
         private Subscription requestContentSubscription;
         private final NettyAdapter adapter;
+        private boolean didEmitHttpResponse;
```
jianghaolu: Does this need to be atomic?

RikkiGibson: I think no, because the purpose of this is for the "happy path", to prevent the .dispose() after completion of the Single from closing the channel before the HttpContents can be read. But we control the timing for that dispose() because we call .onSuccess():

https://github.com/RikkiGibson/autorest-clientruntime-for-java/blob/44be63a9106b7b2b965d70f6341ca64115134df8/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java#L426

jianghaolu: sounds good.

PR #332, opened 2017-12-22
Add a RestProxy.create() overload that takes a ServiceClient parameter

daschult:


RikkiGibson:
Does this streamline gen code?

daschult:
Yeah. A little bit.

PR #337, opened 2018-01-02
Thread safety, native transports, and async file I/O

RikkiGibson:
- Cleans up the stress tests suite: now, stress testing begins by running prepare100MFiles(), then running an upload test, followed by a download test and a download/upload streaming test.

- Some of these test changes revealed thread safety issues which have been addressed with judicious use of volatile, atomics, observeOn(), and assertion that callbacks run on the expected threads.

- Uses reflection to access native epoll and kqueue transports. If users add a native module like netty-transport-native-epoll to their pom, we will use it in NettyClient. We decided to require users to add such a dependency to their pom in order to sidestep build or runtime errors related to unsupported architectures or OSes.

- Zero-copy file I/O is used when using a native transport, passing a FileSegment to upload, and sending to a non-SSL host.

- The file read helper has been rewritten to use callbacks in order to avoid blocking on Future.get(). The helper writeContentToFile(Flowable<byte[]>, AsynchronousFileChannel) has been added for user convenience.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/AsyncInputStream.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/AsyncInputStream.java
@@ -9,7 +9,6 @@ package com.microsoft.rest.v2.http;
 import io.reactivex.Emitter;
 import io.reactivex.Flowable;
 import io.reactivex.functions.BiConsumer;
-import io.reactivex.functions.Function;
 import io.reactivex.internal.util.BackpressureHelper;
 import io.reactivex.schedulers.Schedulers;
 import org.reactivestreams.Subscriber;
@@ -77,29 +76,7 @@ public final class AsyncInputStream {
      * @return The AsyncInputStream.
      */
     public static AsyncInputStream create(final AsynchronousFileChannel fileChannel, final long offset, final long length) {
-        if (true) {
-            return new AsyncInputStream(new FlowableNio2(fileChannel, offset, length), length, true);
-        }
-
-        int numChunks = (int) length / CHUNK_SIZE + (length % CHUNK_SIZE == 0 ? 0 : 1);
-        Flowable<byte[]> fileStream = Flowable.range(0, numChunks).concatMap(new Function<Integer, Flowable<byte[]>>() {
-            ByteBuffer innerBuf = ByteBuffer.wrap(new byte[CHUNK_SIZE]);
-
-            @Override
-            public Flowable<byte[]> apply(Integer chunkNo) throws Exception {
-                final long position = offset + (chunkNo * CHUNK_SIZE);
-                innerBuf.clear();
-                return Flowable.fromFuture(fileChannel.read(innerBuf, position))
-                        .map(new Function<Integer, byte[]>() {
-                            @Override
-                            public byte[] apply(Integer bytesRead) throws Exception {
-                                int bytesWanted = (int) Math.min(offset + length - position, bytesRead);
-                                return Arrays.copyOf(innerBuf.array(), bytesWanted);
-                            }
-                        });
-            }
-        });
-
+        Flowable<byte[]> fileStream = new FileReadFlowable(fileChannel, offset, length);
         return new AsyncInputStream(fileStream, length, true);
     }
 
@@ -114,12 +91,12 @@ public final class AsyncInputStream {
         return create(fileChannel, 0, size);
     }
 
-    private static class FlowableNio2 extends Flowable<byte[]> {
+    private static class FileReadFlowable extends Flowable<byte[]> {
         private final AsynchronousFileChannel fileChannel;
         private final long offset;
         private final long length;
 
-        FlowableNio2(AsynchronousFileChannel fileChannel, long offset, long length) {
+        FileReadFlowable(AsynchronousFileChannel fileChannel, long offset, long length) {
             this.fileChannel = fileChannel;
             this.offset = offset;
             this.length = length;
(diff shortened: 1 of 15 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/http/AsyncInputStream.java
```diff
@@ -104,6 +91,82 @@ public static AsyncInputStream create(AsynchronousFileChannel fileChannel) throw
...
+                    doRead();
+                }
+            }
+
+            void doRead() {
+                innerBuf.clear();
+                fileChannel.read(innerBuf, position, null, onReadComplete);
+            }
+
+            CompletionHandler<Integer, Object> onReadComplete = new CompletionHandler<Integer, Object>() {
```
daschult: Can you mark this as private and final?

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -108,7 +176,12 @@ public void channelCreated(Channel ch) throws Exception {
             }, channelPoolSize);
         }
 
-        private Single<HttpResponse> sendRequestInternalAsync(final HttpRequest request, Proxy proxy) {
+        private boolean useZeroCopy(Channel channel) {
+            boolean isSSL = channel.pipeline().get(SslHandler.class) != null;
+            return !isSSL && !(eventLoopGroup instanceof NioEventLoopGroup);
```
jianghaolu: Can I use an `OioEventLoopGroup`?

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -91,11 +129,41 @@ public void channelCreated(Channel ch) throws Exception {
...
+                    if (cause != null) {
+                        message = cause.getMessage();
+                    }
+                }
+
+                LoggerFactory.getLogger(NettyAdapter.class).debug("Exception when obtaining native EventLoopGroup and SocketChannel: " + message);
+                eventLoopGroup = new NioEventLoopGroup();
+                channelClass = NioSocketChannel.class;
+            }
+            this.eventLoopGroup = eventLoopGroup;
```
daschult: This is identical to the code on line 87. Can you combine these somehow?

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -91,11 +129,41 @@ public void channelCreated(Channel ch) throws Exception {
...
 
+        @SuppressWarnings("unchecked")
         private NettyAdapter(int eventLoopGroupSize, int channelPoolSize) {
-            this.eventLoopGroup = new NioEventLoopGroup(eventLoopGroupSize);
+            MultithreadEventLoopGroup eventLoopGroup;
+            Class<? extends SocketChannel> channelClass;
+            try {
+                final String osName = System.getProperty("os.name");
+                if (osName.contains("Linux")) {
+                    eventLoopGroup = (MultithreadEventLoopGroup) Class.forName("io.netty.channel.epoll.EpollEventLoopGroup").getConstructor(Integer.TYPE).newInstance(eventLoopGroupSize);
```
jianghaolu: Make it more resilient towards maven hiccups? Maybe `NioEventLoopGroup` should be the fallback?

jianghaolu: Ahh I see it down there

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: Ah, yeah--this is really trying to say "if it's a native transport", but if we for some reason moved over to an OioEventLoopGroup, this might do something bad.

Do you think I should check the class's name against a string instead?

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: Writing a line to debug and using NioEventLoopGroup is the intended method to deal with a missing native transport. I guess the only reason to keep checking osName like this is to prevent spurious debug messages ("whoops! Couldn't find an epollEventLoopGroup! I don't care though, I'm on Windows...")

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
jianghaolu: The only 2 we are using are Epoll and KQueue? Can we check if we are using them?

client-runtime/src/main/java/com/microsoft/rest/v2/http/AsyncInputStream.java
RikkiGibson: Good idea.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: I'd like to. The main struggle is the reference to `eventLoopGroup.executorCount()`, which makes it hard to simply delegate the parameterless constructor to a constructor which explicitly takes a pool size and event loop group size.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: Since the classes are not present at compile time, it would be necessary to check with the fully qualified class name, e.g. "io.netty.channel.epoll.EpollEventLoopGroup"

jianghaolu: You are right. I'd prefer that considering there are many event loop group implementations: https://netty.io/4.0/api/io/netty/channel/EventLoopGroup.html

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
daschult: You could change eventLoopGroupSize from `int` to `Integer`. That way if it's null, then you can assume the defaults and only look for the default constructor, not the constructor that takes an `int`.
Also, you should probably add `loadEventLoopGroup(String fullClassName)` and `loadSocketChanellClass(String fullClassName)` methods since your reflection stuff is identical except for the class name.

PR #338, opened 2018-01-03
Add 'innererror' property bag to CloudError contract

demyanenko:
The data contract for ARM errors includes `innererror` field which is a property bag with unspecified contents as per [OData 4.0 spec](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091). This field wasn't used by ARM up until now, but recently there was a change related to Azure Policy that makes heavy use of this property bag. It is already deployed in all regions and consumed by Azure Portal, so now would be a good time to support it in SDKs as well.

The new errors are generated by the ARM Frontdoor itself, so CloudError seems like the most appropriate contract to include it.

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/CloudError.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/CloudError.java
@@ -6,6 +6,8 @@
 
 package com.microsoft.azure;
 
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/CloudError.java
```diff
@@ -33,6 +35,11 @@
      */
     private List<CloudError> details;
 
+    /**
+     * The inner error.
+     */
+    private ObjectNode innererror;
```
jianghaolu: What's the format of the innerError? Will it be a HashMap? (judging on your test)

Also please camel case the variables and methods.

demyanenko:
Looks like all of the issues have been addressed. Is there anything else that needs to be changed?

anuchandy:
@demyanenko based on the discussion with .net runtime team, our understanding is - shape of innerError is not yet finialized with ARM team.   \\cc @shahabhijeet

PR #342, opened 2018-01-09
Use HttpMethod instead of String

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -471,11 +472,11 @@ private boolean isFlowableByteArray(TypeToken entityTypeToken) {
     private Maybe<?> handleBodyReturnTypeAsync(final HttpResponse response, final SwaggerMethodParser methodParser, final Type entityType) {
         final TypeToken entityTypeToken = TypeToken.of(entityType);
         final int responseStatusCode = response.statusCode();
-        final String httpMethod = methodParser.httpMethod();
+        final HttpMethod httpMethod = methodParser.httpMethod();
         final Type returnValueWireType = methodParser.returnValueWireType();
 
         final Maybe<?> asyncResult;
-        if (httpMethod.equalsIgnoreCase("HEAD")
+        if (httpMethod == HttpMethod.HEAD
```
RikkiGibson: So does ExpandableStringEnum use an interned instance when `.fromString()` is called?

client-runtime/src/test/java/com/microsoft/rest/v2/http/HttpMethodTests.java
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    public void HEAD()
+    {
+        assertEquals("HEAD", HttpMethod.HEAD.toString());
+    }
+
+    @Test
+    public void fromStringWithExisting()
+    {
+        assertSame(HttpMethod.POST, HttpMethod.fromString("POST"));
```
RikkiGibson: 👍

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
daschult: Yeah. It checks to see if an instance already exists for the provided String value and will return it if so.

daschult:
CI failure is just a timeout issue. Everything passes locally.

PR #344, opened 2018-01-10
Use URL objects instead of Strings

daschult:


daschult:
This solves https://github.com/Azure/autorest-clientruntime-for-java/issues/340 within the runtime. I'll see if I can fix it in the generator too.

PR #345, opened 2018-01-11
Tweaks for Azure Libraries for Java

RikkiGibson:
- Adds ServiceFuture.fromBody(Single) because it's convenient in a few places in the Fluent core.
- Wraps all checked exceptions thrown from RestProxy.invoke with RuntimeException. I think this is more ergonomic than the java.lang.reflect.UndeclaredThrowableException that results from throwing a checked exception through a proxy method that doesn't have a `throws` clause for that exception.


RikkiGibson:
@daschult This renames pretty much all the RequestPolicies. A corresponding PR is coming to autorest.java to fix test build errors related to that.

PR #346, opened 2018-01-16
Remove check for null body parameter and get tests to pass

daschult:
This solves a problem that Storage raised. They have an API that has an optional body parameter. In the runtime, we were previously always throwing an exception on a null body argument if the method had a parameter with a BodyParam annotation, which means that they were unable to call their API because it would always throw an exception when they passed in null (which is a valid value for an optional parameter).
This is one possible fix for this problem. Another would be to add an argument for BodyParam that indicates whether or the parameter is required. The fix that I chose seemed much easier.


RikkiGibson:
It looks like passing a null body causes no body to get written to the HTTP message on the wire, which I think is preferable to a JSON-encoded null or something like that.

PR #349, opened 2018-01-19
Emit OperationStatus corresponding to Void response

RikkiGibson:
HttpResponse bodies with no content (e.g. most DELETE operations) are represented with a Maybe.empty(), which produces an empty sequence when converted to Observable. This makes it so when you try to get the first response value on a long-running DELETE, you actually make the first GET and return the value from that.

This replaces the value in the Maybe.empty() with an OperationStatus with a null value inside so that the unnecessary GET doesn't occur.

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
@@ -32,6 +32,7 @@ import com.microsoft.rest.v2.http.HttpResponse;
 import io.reactivex.Observable;
 import io.reactivex.ObservableSource;
 import io.reactivex.Single;
+import io.reactivex.SingleSource;
 import io.reactivex.exceptions.Exceptions;
 import io.reactivex.functions.Function;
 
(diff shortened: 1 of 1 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -256,14 +258,20 @@ protected Object handleAsyncHttpResponse(final HttpRequest httpRequest, Single<H
...
+                                            }
+                                        }).switchIfEmpty(Single.defer(new Callable<SingleSource<? extends OperationStatus<?>>>() {
+                                            @Override
+                                            public SingleSource<? extends OperationStatus<?>> call() throws Exception {
+                                                return Single.just(new OperationStatus<>((Object) null, pollStrategy.status()));
+                                            }
+                                        })).toObservable();
                                 Observable<OperationStatus<Object>> rest = pollStrategy.pollUntilDoneWithStatusUpdates(httpRequest, methodParser, operationStatusResultType);
-                                return first.concatWith(rest);
+                                return first.concatWith(first).concatWith(rest);
```
RikkiGibson: Oops--weird that I somehow left this in there.

PR #351, opened 2018-01-23
 applicationCredentials is set to null even when passed

### LINKED ISSUE #350: applicationCredentials is set to null even when passed.
Currently,  when 
`DelegatedTokenCredentials(ApplicationTokenCredentials applicationCredentials, String redirectUrl, String authorizationCode)`
 here:[Constructor](https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/DelegatedTokenCredentials.java#L62) is called, the applicationCredentials is not set.

This is the missing line : https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/DelegatedTokenCredentials.java#L49

Due to the applicationCredentials not being present we get a null pointer : https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/DelegatedTokenCredentials.java#L169

The `DelegatedTokenCredentials(ApplicationTokenCredentials applicationCredentials, String redirectUrl, String authorizationCode)` constructor should ideally set authorization code and then call : `DelegatedTokenCredentials(ApplicationTokenCredentials applicationCredentials, String redirectUrl)`

spacharya:
fixes #350 

Signed-off-by: Suraj Acharya <suraj.spa@gmail.com>

### DIFF
--- a/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/DelegatedTokenCredentials.java
+++ b/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/DelegatedTokenCredentials.java
@@ -60,11 +60,8 @@ public class DelegatedTokenCredentials extends AzureTokenCredentials {
      * @param authorizationCode the oauth2 authorization code
      */
     public DelegatedTokenCredentials(ApplicationTokenCredentials applicationCredentials, String redirectUrl, String authorizationCode) {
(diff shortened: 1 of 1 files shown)


jianghaolu:
Nice catch! Thanks for contributing! 🎉

PR #354, opened 2018-01-23
Use Flowable<byte[]> instead of AsyncInputStream. Drop File body types.

RikkiGibson:
After this morning's discussion, we realized that Flowable<byte[]> was a better general use stream type and wrapping it with AsyncInputStream is not necessary.

Some refactoring related to how Content-Length is applied to requests conflicted with the way FileSegment worked, and since it's become pretty clear that we're not as interested in FileSegment due to being harder to compose than a Flowable<byte[]> and are not going to ship zero-copy uploads due to the limited and finicky nature of it, I deleted the File-specific body types.


client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -39,6 +36,7 @@
 import io.reactivex.exceptions.Exceptions;
 import io.reactivex.functions.Function;
 import org.joda.time.DateTime;
+import org.reactivestreams.Publisher;
```
daschult: Is this a new dependency? I've never seen the `org.reactivestreams` package before.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpRequest.java
```diff
@@ -171,7 +172,6 @@ public HttpRequest withBody(byte[] body, String mimeContentType) {
      */
     public HttpRequest withBody(HttpRequestBody body) {
```
daschult: Since this method doesn't update the content length header, it seems like this could be a dangerous method for a customer to call. I'd suggest making it private or at least package private.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
RikkiGibson: org.reactivestreams is the Reactive Streams spec that Java 9 Flows adhere to, and RxJava2 references and adheres to. The entire spec is just a few interfaces that specify a very small subset of what RxJava2 and Flowable in particular can do.

For some reason a few RxJava methods like flatMapPublisher take a Function<T, Publisher<U>>, but themselves return a Flowable<U>, instead of being called flatMapFlowable.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpRequest.java
RikkiGibson: I kept tinkering with this idea, thinking about what RequestPolicy use case could exist for calling this method, about what we could do with the docs or parameters to indicate "hey! you might need to set the content length!"

Then I tried making it package-private and I wasn't able to access it in RestProxy.java. So it goes.

RikkiGibson:
Removed HttpRequestBody.
- I want to make sure there's a single source of truth for headers like Content-Length and Content-Type as much as possible. These perhaps could be convenience methods on the HttpHeaders object but storing them independently on the body makes it easy for them to go out of sync. withBody() was already failing to set the Content-Type even though it was being passed as a parameter.
- Flowable<byte[]> is looking like our common body content interface and given my previous point it seems adequate as a body type.

PR #360, opened 2018-02-02
Perform response deserialization in HttpPipeline

RikkiGibson:
- SerializerAdapter.Encoding is now SerializerEncoding
- Adds `Object deserializedHeaders` and `Object deserializedBody` members to HttpResponse
- Adds HttpResponseDecoder which needs to get attached to the HttpResponse in order for decoding to work. I'm slightly concerned about using `withResponseDecoder()` instead of requiring a constructor parameter. It's just a lot of code churn if I add a constructor parameter. Let me know what you think.
- Removes `bodyAsInputStreamAsync()` from HttpResponse
  - We were never actually streaming in impls of this and it doesn't seem worth it at this time to provide an alternative mechanism for streaming.
  - If we bring it back down the line (say in order to let people implement other HttpClients more faithfully) then I think it should return simply InputStream instead of Single<InputStream> because just obtaining the InputStream shouldn't be waiting for async reads to take place
- HttpResponse and RestResponse are now Closeable
  - These close methods only have an effect if a streaming body is present

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
@@ -68,65 +68,48 @@ public final class AzureAsyncOperationPollStrategy extends PollStrategy {
     }
 
     @Override
-    public Single<HttpResponse> updateFromAsync(HttpResponse httpPollResponse) {
-        return ensureExpectedStatus(httpPollResponse)
-                .flatMap(new Function<HttpResponse, Single<HttpResponse>>() {
-                    @Override
-                    public Single<HttpResponse> apply(HttpResponse response) {
-                        updateDelayInMillisecondsFrom(response);
+    public Single<HttpResponse> updateFromAsync(final HttpResponse httpPollResponse) {
+        updateDelayInMillisecondsFrom(httpPollResponse);
 
-                        Single<HttpResponse> result;
-                        if (!pollingCompleted) {
-                            final HttpResponse bufferedHttpPollResponse = response.buffer();
-                            result = bufferedHttpPollResponse.bodyAsStringAsync()
-                                    .map(new Function<String, HttpResponse>() {
-                                        @Override
-                                        public HttpResponse apply(String bodyString) {
-                                            AsyncOperationResource operationResource = null;
+        if (!pollingCompleted) {
+            AsyncOperationResource operationResource = null;
 
-                                            try {
-                                                operationResource = deserialize(bodyString, AsyncOperationResource.class);
-                                            }
-                                            catch (IOException ignored) {
-                                            }
+            try {
+                operationResource = reserialize(httpPollResponse.deserializedBody(), AsyncOperationResource.class);
+            }
+            catch (IOException ignored) {
+            }
 
-                                            if (operationResource == null || operationResource.status() == null) {
-                                                throw new CloudException("The polling response does not contain a valid body", bufferedHttpPollResponse, null);
-                                            }
-                                            else {
-                                                final String status = operationResource.status();
-                                                setStatus(status);
+            if (operationResource == null || operationResource.status() == null) {
+                throw new CloudException("The polling response does not contain a valid body", httpPollResponse, null);
+            }
+            else {
+                final String status = operationResource.status();
+                setStatus(status);
 
-                                                pollingCompleted = OperationState.isCompleted(status);
-                                                if (pollingCompleted) {
-                                                    pollingSucceeded = OperationState.SUCCEEDED.equalsIgnoreCase(status);
-                                                    clearDelayInMilliseconds();
+                pollingCompleted = OperationState.isCompleted(status);
+                if (pollingCompleted) {
+                    pollingSucceeded = OperationState.SUCCEEDED.equalsIgnoreCase(status);
+                    clearDelayInMilliseconds();
 
-                                                    if (!pollingSucceeded) {
-                                                        throw new CloudException("Async operation failed with provisioning state: " + status, bufferedHttpPollResponse);
-                                                    }
-
-                                                    if (operationResource.id() != null) {
-                                                        gotResourceResponse = true;
-                                                    }
-                                                }
-                                            }
-
-                                            return bufferedHttpPollResponse;
-                                        }
-                                    });
-                        }
-                        else {
-                            if (pollingSucceeded) {
-                                gotResourceResponse = true;
-                            }
-
-                            result = Single.just(response);
-                        }
-
-                        return result;
+                    if (!pollingSucceeded) {
+                        throw new CloudException("Async operation failed with provisioning state: " + status, httpPollResponse);
                     }
-                });
+
+                    if (operationResource.id() != null) {
+                        gotResourceResponse = true;
+                    }
+                }
+            }
+        }
+        else {
+            if (pollingSucceeded) {
+                gotResourceResponse = true;
+            }
+
+        }
+
+        return Single.just(httpPollResponse);
     }
 
     @Override
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
@@ -64,11 +64,11 @@ public final class AzureAsyncOperationPollStrategy extends PollStrategy {
             throw new IllegalStateException("Polling is completed and did not succeed. Cannot create a polling request.");
         }
 
-        return new HttpRequest(fullyQualifiedMethodName(), HttpMethod.GET, pollUrl);
+        return new HttpRequest(fullyQualifiedMethodName(), HttpMethod.GET, pollUrl, createResponseDecoder());
     }
 
     @Override
-    public Single<HttpResponse> updateFromAsync(final HttpResponse httpPollResponse) {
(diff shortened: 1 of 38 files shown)


azure-client-runtime/src/main/java/com/microsoft/azure/v2/ProvisioningStatePollStrategy.java
```diff
@@ -48,31 +48,26 @@ HttpRequest createPollRequest() {
                                 .map(new Function<String, HttpResponse>() {
                                     @Override
                                     public HttpResponse apply(String responseBody) {
-                                            ResourceWithProvisioningState resource = null;
```
RikkiGibson: Looks like here the indent depth just got changed by IntelliJ.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureProxy.java
```diff
@@ -299,7 +300,7 @@ protected Object handleAsyncHttpResponse(final HttpRequest httpRequest, Single<H
                     public Single<PollStrategy> apply(final HttpResponse originalHttpResponse) {
                         final int httpStatusCode = originalHttpResponse.statusCode();
                         final int[] longRunningOperationStatusCodes = new int[] {200, 201, 202};
-                        return ensureExpectedStatus(originalHttpResponse, methodParser, longRunningOperationStatusCodes)
```
RikkiGibson: This change looks like a mistake.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/CompletedPollStrategy.java
```diff
@@ -20,30 +20,29 @@
  * further polling.
  */
 public class CompletedPollStrategy extends PollStrategy {
-    private final HttpResponse bufferedOriginalHttpResponse;
```
RikkiGibson: If it was important to say in the variable's name that it's buffered it seems worth it to express in the static type system that it's buffered. Might change this. A lot of these little changes in LROs were due to just clumsily trying to get it all working again after moving deserialization into HttpResponseDecoder.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/PollStrategy.java
```diff
@@ -148,10 +150,25 @@ void setStatus(String status) {
...
                             @Override
                             public Single<HttpResponse> call() throws Exception {
-                                final HttpRequest pollRequest = createPollRequest();
+                                final HttpRequest pollRequest = createPollRequest()
+                                        .withResponseDecoder(new HttpResponseDecoder(methodParser, restProxy.serializer()));
+
                                 return restProxy.sendHttpRequestAsync(pollRequest);
                             }
                         }))
+                        .onErrorResumeNext(new Function<Throwable, SingleSource<? extends HttpResponse>>() {
```
RikkiGibson: This shouldn't be necessary.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/CompletedPollStrategy.java
```diff
@@ -20,30 +21,29 @@
...
      *                     initiated the long running operation.
-     * @param originalHttpResponse The HTTP response to the original HTTP request.
+     * @param firstHttpResponse The HTTP response to the original HTTP request.
      */
-    public CompletedPollStrategy(RestProxy restProxy, SwaggerMethodParser methodParser, HttpResponse originalHttpResponse) {
+    public CompletedPollStrategy(RestProxy restProxy, SwaggerMethodParser methodParser, HttpResponse firstHttpResponse) {
         super(restProxy, methodParser, 0);
-
-        this.bufferedOriginalHttpResponse = originalHttpResponse.buffer();
+        this.firstHttpResponse = firstHttpResponse.buffer();
```
RikkiGibson: This isn't actually sound if the httpResponse passed were not already buffered--will probably fix this up when it comes time to revisit LROs.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -356,32 +228,19 @@ private Exception instantiateUnexpectedException(SwaggerMethodParser methodParse
...
-                    || contentType.startsWith("application/xml")
-                    || contentType.startsWith("text/xml");
-
-            final Object exceptionBody = responseContent.isEmpty() || !isSerializableContentType
-                    ? null
-                    : serializer.deserialize(responseContent, exceptionBodyType, bodyEncoding(response.headers()));
-
-            result = exceptionConstructor.newInstance("Status code " + responseStatusCode + ", " + bodyRepresentation, response, exceptionBody);
-        } catch (IllegalAccessException | InstantiationException | InvocationTargetException | NoSuchMethodException | JsonParseException e) {
+            result = exceptionConstructor.newInstance("Status code " + responseStatusCode + ", " + bodyRepresentation, response, response.deserializedBody());
```
RikkiGibson: Slight behavior change: it's no longer so easy to distinguish whether the error body was absent or it failed to deserialize. I'd have to save the JsonParsingException somewhere when decoding. So now I'm returning RestException if the error body failed to deserialize, but at least the raw content will still be visible.

client-runtime/src/main/java/com/microsoft/rest/v2/RestProxy.java
```diff
@@ -560,7 +401,7 @@ else if (returnTypeToken.isSubtypeOf(Single.class)) {
                 public Single<?> apply(HttpResponse response) throws Exception {
                     return handleRestResponseReturnTypeAsync(response, methodParser, singleTypeParam);
                 }
-            });
+            }).onErrorResumeNext(WARN_MISSING_DECODING);
```
RikkiGibson: This slightly strange way of adding the missing decoder warning is due to the fact that certain methods of composition don't work with wildcards.

azure-client-runtime/src/main/java/com/microsoft/azure/v2/AzureAsyncOperationPollStrategy.java
```diff
@@ -68,7 +68,7 @@ else if (pollingSucceeded) {
     }
 
     @Override
-    public Single<HttpResponse> updateFromAsync(HttpResponse httpPollResponse) {
+    public Single<HttpResponse> updateFromAsync(final HttpResponse httpPollResponse) {
```
daschult: This `final` modifier doesn't really seem necessary.

client-runtime/src/main/java/com/microsoft/rest/v2/protocol/HttpResponseDecoder.java
```diff
@@ -0,0 +1,291 @@
...
+
+        if (token.isSubtypeOf(Single.class) || token.isSubtypeOf(Maybe.class) || token.isSubtypeOf(Observable.class)) {
+            token = TypeToken.of(getTypeArgument(token.getType()));
+        }
+
+        if (token.isSubtypeOf(RestResponse.class)) {
+            token = TypeToken.of(getTypeArguments(token.getType())[1]);
+        }
+
+        // TODO: unwrap OperationStatus a different way?
```
RikkiGibson: Not extremely happy with this but also reluctant to introduce polymorphism (i.e. attaching an AzureResponseDecoder to the HttpRequest) until LROs are refactored. It might wind up being the best way though.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpRequest.java
```diff
@@ -18,6 +19,7 @@
     private String callerMethod;
     private HttpMethod httpMethod;
     private URL url;
+    private HttpResponseDecoder responseDecoder;
```
daschult: Is there any reason a `responseDecoder` would change? Can we make this a constructor parameter and set it to final?

client-runtime/src/main/java/com/microsoft/rest/v2/protocol/HttpResponseDecoder.java
```diff
@@ -0,0 +1,291 @@
...
+            && !entityTypeToken.isSubtypeOf(byte[].class)
+            && !entityTypeToken.isSubtypeOf(Boolean.TYPE) && !entityTypeToken.isSubtypeOf(Boolean.class)
+            && !entityTypeToken.isSubtypeOf(Void.TYPE) && !entityTypeToken.isSubtypeOf(Void.class);
+
+        int[] expectedStatuses = methodParser.expectedStatusCodes();
+        boolean isErrorStatus = true;
+        if (expectedStatuses != null) {
+            for (int i = 0; i < expectedStatuses.length && isErrorStatus; i++) {
+                if (expectedStatuses[i] == response.statusCode()) {
+                    isErrorStatus = false;
```
daschult: Don't forget to `break` after the condition you want is encountered. It makes things a little quicker.

client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpRequest.java
RikkiGibson: Yeah, I've been thinking `final` might be best here--just have to deal with the churn in all the places where an HttpRequest was created.

client-runtime/src/main/java/com/microsoft/rest/v2/protocol/HttpResponseDecoder.java
RikkiGibson: This makes me think that simply putting a `break;` is going to be clearer to maintainers than `&& isErrorStatus` in the loop condition.

Issue #357, opened 2018-01-31
Use Flowable<ByteBuffer> for streaming


RikkiGibson:
After much discussion and weighing of alternatives, we've settled on the usage of Flowable streams which produce *unpooled* ByteBuffers. This allows use cases like memory mapped files and uploading views of underlying buffers without copying.

PR #361, opened 2018-02-02
Use Flowable<ByteBuffer> instead of Flowable<byte[]>

RikkiGibson:
After considering the merits of a significant number of potential designs, it appears we've settled on "one true streaming type". ByteBuffer lets us pass around "views" of existing memory instead of needing to copy into a byte[] of exactly the right size. It allows users to upload from direct memory if they desire as well as memory mapped files.

- Factored the netty version out into a maven property
  - We will want to update to 4.1.21.Final whenever it comes out to get a [particular bug fix](https://github.com/netty/netty/issues/7672).
- Added a memory mapped file upload test
- Added `FlowableUtil.split(ByteBuffer, chunkSize)`
  - This is important when combined with SSL because (it seems) the SSLHandler will try to encode the entirety of each message you send before writing it--if you don't split large messages before writing them to the channel you'll run out of direct memory.


RikkiGibson:
Closes #357.

PR #365, opened 2018-02-05
Add ByteBufferUtil and tests

daschult:
Add a ByteBufferUtil class that can convert from ByteBuffers to byte[]s.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/util/ByteBufferUtil.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/util/ByteBufferUtil.java
@@ -13,9 +13,9 @@ import java.nio.ByteBuffer;
  */
 public class ByteBufferUtil {
     /**
-     * Convert the provided ByteBuffer to a byte[].
-     * @param byteBuffer The ByteBuffer to convert.
-     * @return The converted byte[].
+     * Read the contents of the ByteBuffer into a byte[].
+     * @param byteBuffer The ByteBuffer to read from.
+     * @return The created byte[].
      */
     public static byte[] toByteArray(ByteBuffer byteBuffer) {
         byte[] byteArray = null;
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/util/ByteBufferUtil.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/util/ByteBufferUtil.java
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.rest.v2.util;
+
+import java.nio.ByteBuffer;
+
+/**
+ * Contains helper methods for dealing with Flowables.
+ */
+public class ByteBufferUtil {
+    /**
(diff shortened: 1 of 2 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/util/ByteBufferUtil.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Contains helper methods for dealing with Flowables.
+ */
+public class ByteBufferUtil {
+    /**
+     * Convert the provided ByteBuffer to a byte[].
+     * @param byteBuffer The ByteBuffer to convert.
+     * @return The converted byte[].
+     */
+    public static byte[] toByteArray(ByteBuffer byteBuffer) {
```
RikkiGibson: Do you think it would be useful to somehow convey that this is a copying operation and not a reference to the underlying `.array()`?

client-runtime/src/test/java/com/microsoft/rest/v2/util/ByteBufferUtilTests.java
```diff
@@ -0,0 +1,30 @@
...
+        assertArrayEquals(new byte[0], ByteBufferUtil.toByteArray(ByteBuffer.wrap(new byte[0])));
+    }
+
+    @Test
+    public void toByteArrayWithNonEmptyByteBuffer()
+    {
+        final ByteBuffer byteBuffer = ByteBuffer.wrap(new byte[] { 0, 1, 2, 3, 4 });
+        assertEquals(5, byteBuffer.remaining());
+        assertArrayEquals(new byte[] { 0, 1, 2, 3, 4 }, ByteBufferUtil.toByteArray(byteBuffer));
+        assertEquals(0, byteBuffer.remaining());
```
RikkiGibson: Depending on the behavior you settle on it might be reasonable to assert that the return value *is not* the same byte[] instance as the .array() on the byteBuffer.

client-runtime/src/main/java/com/microsoft/rest/v2/util/ByteBufferUtil.java
daschult: I updated the comment to say that it is a read operation.

client-runtime/src/test/java/com/microsoft/rest/v2/util/ByteBufferUtilTests.java
daschult: Done.

RikkiGibson:
👍

client-runtime/src/main/java/com/microsoft/rest/v2/util/ByteBufferUtil.java
```diff
@@ -0,0 +1,28 @@
...
+     * @return The created byte[].
+     */
+    public static byte[] toByteArray(ByteBuffer byteBuffer) {
+        byte[] byteArray = null;
+        if (byteBuffer != null) {
+            byteArray = new byte[byteBuffer.remaining()];
+            byteBuffer.get(byteArray);
+        }
+        return byteArray;
+    }
```
RikkiGibson: I forgot to mention this, but next time you touch this class you might want to add a private constructor. Easy to forget and not likely to result in misuse, so it's no big deal.

PR #366, opened 2018-02-05
Add FlowableUtil.collectBytesInBuffer()

daschult:
There is a FlowableUtil.collectBytes() method that produces a `Single<byte[]>`, but we're transitioning over to using `ByteBuffer` instead of `byte[]`, so I renamed that method and created a new one that returns a `Single<ByteBuffer>`.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
@@ -55,10 +55,10 @@ public final class FlowableUtil {
 
     /**
      * Collects byte buffers emitted by a Flowable into a byte array.
-     * @param content A stream which emits byte arrays.
+     * @param content A stream which emits byte buffers.
      * @return A Single which emits the concatenation of all the byte buffers given by the source Flowable.
      */
-    public static Single<byte[]> collectBytes(Flowable<ByteBuffer> content) {
+    public static Single<byte[]> collectBytesInArray(Flowable<ByteBuffer> content) {
         return content.collectInto(ByteStreams.newDataOutput(), new BiConsumer<ByteArrayDataOutput, ByteBuffer>() {
             @Override
             public void accept(ByteArrayDataOutput out, ByteBuffer chunk) throws Exception {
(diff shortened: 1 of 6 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
```diff
@@ -55,10 +55,10 @@ public static boolean isFlowableByteBuffer(TypeToken entityTypeToken) {
 
     /**
      * Collects byte buffers emitted by a Flowable into a byte array.
-     * @param content A stream which emits byte arrays.
```
RikkiGibson: Good catch on the documentation rot..

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
```diff
@@ -55,10 +55,10 @@ public static boolean isFlowableByteBuffer(TypeToken entityTypeToken) {
...
+     * @param content A stream which emits byte buffers.
      * @return A Single which emits the concatenation of all the byte buffers given by the source Flowable.
      */
-    public static Single<byte[]> collectBytes(Flowable<ByteBuffer> content) {
+    public static Single<byte[]> collectBytesInArray(Flowable<ByteBuffer> content) {
         return content.collectInto(ByteStreams.newDataOutput(), new BiConsumer<ByteArrayDataOutput, ByteBuffer>() {
             @Override
             public void accept(ByteArrayDataOutput out, ByteBuffer chunk) throws Exception {
```
RikkiGibson: BTW if you have any idea on how to do this elegantly that reduces copying, that would be welcome :)

Maybe something that just holds references to all the ByteBuffers, takes the sum of their `.remaining()` and then copies it all into one byte array one by one. That's probably what Netty is doing for us [here](https://github.com/Azure/autorest-clientruntime-for-java/blob/v2/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyResponse.java#L53).

RikkiGibson:
There will be breaks to fix in autorest.java, but nothing a little find and replace can't handle.

Issue #362, opened 2018-02-02
Native SSL with Netty


RikkiGibson:
http://netty.io/wiki/forked-tomcat-native.html

Netty offers several optional modules which use native libraries for SSL. I've been told by others who've been using Netty that this can improve performance significantly.

This is perhaps an opportunity to solidify our stance on native transports as well as native SSL. It would be nice to just always do the best thing for our users with regard to this. However there are so many ways to obtain native SSL, and there are some bugs that come up that are specific to native transports.

I think the most realistic way forward is probably to just show users what they can put in their pom to cause the appropriate native modules to be used, and that way they can test for themselves if the native modules work in their environment.

Alternatively we might be able to default to automatically including the native modules that are simplest to incorporate and add some build-time property to disable them if users want to do their own customization. Adding a "pom-only" project that they can depend on (client-runtime-native/pom.xml)? might even work.

RikkiGibson:
Let me know the perspective on this on your side @rickle-msft

RikkiGibson:
Addressed by #369

PR #369, opened 2018-02-06
Add client-runtime-native module

RikkiGibson:
This is my proposal for #362. By default, people who grab our runtime and libraries that depend on our runtime will use the transport and SSL based on the Java standard library. But if they want the increased performance (and potential bugs) they can add this dependency:

```xml
<dependency>
  <groupId>com.microsoft.rest.v2</groupId>
  <artifactId>client-runtime-native</artifactId>
  <version>2.0.0-SNAPSHOT</version>
  <type>pom</type>
</dependency>
```

Then each SDK's README should show this dependency in the "how to install" section next to the SDK dependency itself.

This will add either epoll or kqueue transport if applicable, and will also add a statically-linked boringssl dependency on windows/linux/mac. Users who want fine grained control over the SSL implementation used (e.g., I know we'll have OpenSSL already installed on the server, I want to use that, etc.) would not use client-runtime-native and would instead just add the appropriate native dependencies individually themselves.

Instead of needing to know which Netty version is being used in the runtime, they'll need to know what version of the runtime is being used by the particular Azure SDK-- so it's not perfect, but it's probably an improvement over making them know the right versions of Netty native modules.

### DIFF
--- a/client-runtime-native/pom.xml
+++ b/client-runtime-native/pom.xml
@@ -20,16 +20,6 @@
     </description>
     <url>https://github.com/Azure/autorest-clientruntime-for-java</url>
 
-    <build>
-        <extensions>
-            <extension>
-                <groupId>kr.motd.maven</groupId>
-                <artifactId>os-maven-plugin</artifactId>
-                <version>1.5.0.Final</version>
-            </extension>
-        </extensions>
-    </build>
-
     <profiles>
         <profile>
             <id>osx</id>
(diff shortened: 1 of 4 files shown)


RikkiGibson:
I still need to test and make sure this particular setup does the right thing on Windows and Linux.

RikkiGibson:
Working on macOS, Linux and Windows 👍

PR #371, opened 2018-02-07
Add support for x-ms-header-collection-prefix extension

daschult:
This adds support for writing x-ms-header-collection-prefix header parameters. It doesn't support reading x-ms-header-collection-prefix headers from a response. That needs to be done in a separate pull request.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodParser.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodParser.java
@@ -291,9 +291,9 @@ public class SwaggerMethodParser {
                 if (0 <= parameterIndex && parameterIndex < swaggerMethodArguments.length) {
                     final Object methodArgument = swaggerMethodArguments[headerSubstitution.methodParameterIndex()];
                     if (methodArgument instanceof Map) {
-                        final Map<String,?> headerCollection = (Map<String,?>)methodArgument;
+                        final Map<String, ?> headerCollection = (Map<String, ?>) methodArgument;
                         final String headerCollectionPrefix = headerSubstitution.urlParameterName();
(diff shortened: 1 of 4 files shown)


RikkiGibson:
So is this basically expanding the purpose of the HeaderParam so that if you provide a Map as the argument type it uses the value for the HeaderParam as a prefix? I think that's fine, but it would be good to expand the doc for HeaderParam to explain this.

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodParser.java
```diff
@@ -265,9 +285,26 @@ public String path(Object[] methodArguments) {
...
 
-        final Iterable<EncodedParameter> substitutedHeaders = getEncodedParameters(headerSubstitutions, swaggerMethodArguments, null);
-        for (final EncodedParameter substitutedHeader : substitutedHeaders) {
-            result.set(substitutedHeader.name(), substitutedHeader.encodedValue());
+        if (headerSubstitutions != null) {
+            for (Substitution headerSubstitution : headerSubstitutions) {
+                final int parameterIndex = headerSubstitution.methodParameterIndex();
+                if (0 <= parameterIndex && parameterIndex < swaggerMethodArguments.length) {
+                    final Object methodArgument = swaggerMethodArguments[headerSubstitution.methodParameterIndex()];
+                    if (methodArgument instanceof Map) {
```
RikkiGibson: If you want, you can get the full static type information from methodParser

RikkiGibson: Strictly speaking, the keys just need to be something that converts gracefully to String.

RikkiGibson:
Just need checkstyle fixes.

PR #373, opened 2018-02-08
Added option to set SSLSocketFactory on ApplicationTokenCredentials

gitorko:
Truststore with certs loaded at runtime fail if the SSLSocketFactory is not set.

We use a custom truststore for certs, The certs are loaded at runtime and even though we pass the RestClient with the OkHttpClient and SSLSocketFactory the certs loaded at runtime are not identified.

We get the error:
javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)

Since SslSocketFactory was never passed to AuthenticationContext it was not able to detect the certs loaded at runtime. With the change above we can now pass SslSocketFactory at runtime.

ApplicationTokenCredentials credentials = new ApplicationTokenCredentials(connectionInfo.getClientId(), connectionInfo.getTenantId(), connectionInfo.getClientSecret(), AzureEnvironment.AZURE).withSSLSocketFactory(getYourOwnCustomSSLSocketFactory());

### DIFF
--- a/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/ApplicationTokenCredentials.java
+++ b/azure-client-authentication/src/main/java/com/microsoft/azure/credentials/ApplicationTokenCredentials.java
@@ -18,6 +18,7 @@ import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.Proxy;
 import java.security.KeyFactory;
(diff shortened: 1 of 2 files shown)


jianghaolu:
Cool. Thanks for contributing!

PR #376, opened 2018-02-09
Add runtime support for deserializing HeaderCollections

daschult:


client-runtime/src/main/java/com/microsoft/rest/v2/protocol/HttpResponseDecoder.java
```diff
@@ -292,6 +296,50 @@ private Object deserializeHeaders(HttpHeaders headers) throws IOException {
...
+                                            headerCollection.put(headerName.substring(headerCollectionPrefixLength), header.value());
+                                        }
+                                    }
+
+                                    final boolean declaredFieldAccessibleBackup = declaredField.isAccessible();
+                                    try {
+                                        if (!declaredFieldAccessibleBackup) {
+                                            declaredField.setAccessible(true);
+                                        }
+                                        declaredField.set(deserializedHeaders, headerCollection);
```
RikkiGibson: Ah, so when we serialize/deserialize we take care of all the fixed-size fields? And then we use reflection to set the HeaderCollection property?

daschult: Correct.

daschult: I'll add the HeaderCollection annotation in the generator.

RikkiGibson:
Just Checkstyle failures

Issue #372, opened 2018-02-07
Graceful teardown of HttpClient


RikkiGibson:
Currently a NettyAdapter is associated with a SharedChannelPool and EventLoopGroup. It's reasonable that people would want to tear it down and free the resources. Therefore HttpClient.Factory should support some kind of disposal. A few questions should then be answered:

- Does anything need to change about convenience methods which create HttpClients implicitly?
  - Might be simplest to just make a shared static default HttpClient.Factory that we don't tear down? If we do this, it might be good to try and make it scale itself down when it's idle
  - HttpClient or HttpPipeline could be made Disposable? This could be reasonable in cases where we know that nobody but that particular HttpClient or HttpPipeline ever used a particular HttpClient.Factory and would spare users from needing to hold on to an HttpClient.Factory separately themselves.

SharedChannelPool has some logic related to shutdown but I don't think it's been exercised at this point. We'll also want to determine whether shutting down will immediately close all connections and end the subscriptions with an error or if it will cause the SharedChannelPool to enter some kind of "graceful closing" mode where it services the existing connections to completion but refuses any further requests. Consider `ExecutorService.shutdown()` versus `ExecutorService.shutdownNow()` as a model.

RikkiGibson:
This is pretty much done, will just get around to adding a synchronous overload of `shutdown()` so we can conform to Closeable

PR #380, opened 2018-02-14
Add shutdown() method to HttpClient.Factory

RikkiGibson:
Users should be able to free the resources associated with an HttpClient when they're done using it. This PR makes it so:
- All default HTTP pipelines share a static, lazy instantiated HttpClientFactory so that we don't leak the implicitly created resources (EventLoopGroups, channels)
- Manually created HttpPipelines require users to hold on to the HttpClientFactory and close it if they want to reclaim the resources
  - The only alternative I can really think of for this is to add shutdown() methods to HttpPipeline and/or HttpClient themselves and to put a ref count on the HttpClient.Factory-- seems overly complicated

Addresses #372.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -66,7 +66,7 @@ public final class NettyClient extends HttpClient {
      * @param configuration the HTTP client configuration.
      * @param adapter the adapter to Netty
      */
-    private NettyClient(HttpClient.Configuration configuration, NettyAdapter adapter) {
+    private NettyClient(HttpClientConfiguration configuration, NettyAdapter adapter) {
         this.adapter = adapter;
         this.proxy = configuration == null ? null : configuration.proxy();
     }
@@ -567,7 +567,7 @@ public final class NettyClient extends HttpClient {
     /**
      * The factory for creating a NettyClient.
      */
-    public static class Factory implements HttpClient.Factory {
+    public static class Factory implements HttpClientFactory {
         private final NettyAdapter adapter;
 
         /**
@@ -588,7 +588,7 @@ public final class NettyClient extends HttpClient {
         }
 
         @Override
-        public HttpClient create(final Configuration configuration) {
+        public HttpClient create(final HttpClientConfiguration configuration) {
             return new NettyClient(configuration, adapter);
         }
 
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -274,12 +274,8 @@ public final class NettyClient extends HttpClient {
                                         }
                                     }
                                 });
-                            } catch (Exception e) {
-                                emitErrorIfSubscribed(e);
-                            }
 
-                            if (request.body() == null) {
-                                try {
+                                if (request.body() == null) {
                                     channel.writeAndFlush(DefaultLastHttpContent.EMPTY_LAST_CONTENT)
                                             .addListener(new GenericFutureListener<Future<? super Void>>() {
                                                 @Override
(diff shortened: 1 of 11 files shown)


RikkiGibson:
Also went ahead and fixed a leak where cancelling a `.toList()` subscription on a `Flowable<ByteBuf>` would result in all the ByteBufs accumulated in the list being dropped without being released. I'm not getting rid of `Flowable<ByteBuf>` entirely but we don't want users to have access to the type and we have to be careful which operators we apply in case some of them cause values to be dropped, due to cancellation, error or otherwise.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -258,31 +264,38 @@ public boolean isDisposed() {
...
+                                        }
                                     }
-                                }
-                            });
+                                });
+                            } catch (Exception e) {
+                                emitErrorIfSubscribed(e);
+                            }
 
                             if (request.body() == null) {
```
daschult: Is this section still valid if an exception is caught on lines 277 - 278?

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -309,8 +322,12 @@ public void onNext(ByteBuffer buf) {
...
-                                        channel.writeAndFlush(Unpooled.wrappedBuffer(buf))
-                                                .addListener(onChannelWriteComplete);
+                                        try {
+                                            channel.writeAndFlush(Unpooled.wrappedBuffer(buf))
+                                                    .addListener(onChannelWriteComplete);
+                                        } catch (Exception e) {
+                                            emitErrorIfSubscribed(e);
+                                        }
 
                                         if (channel.isWritable()) {
```
daschult: Is this section still valid if an exception is caught on lines 328-329?

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyResponse.java
```diff
@@ -64,19 +74,12 @@ public ByteBuf apply(List<ByteBuf> l) {
     public Single<byte[]> bodyAsByteArrayAsync() {
         return collectContent().map(new Function<ByteBuf, byte[]>() {
             @Override
-            public byte[] apply(ByteBuf byteBuf) {
-                return toByteArray(byteBuf);
+            public byte[] apply(ByteBuf byteBuf) throws Exception {
+                return byteBuf.array();
```
daschult: Doesn't this function expose the internal array of the ByteBuf? I know that you've done a lot to avoid copying bits, but I just want to make sure you're confident that you want to expose the internal array of a ByteBuf.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: No I don't think so-- actually, it might be easiest to just throw here. The reason these try/catch blocks were added is that addListener will fail after the EventLoopGroup is shut down.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyResponse.java
RikkiGibson: It does. The reason that's ok is that the ByteBuf received by this method comes from `Unpooled.buffer` on line 59. Actually, I would expect a buffer from the pool to throw because there's only direct memory backing it and not an array.

It might be worth playing with how we aggregate the ByteBuf content for a more efficient conversion to byte[] or String (probably String is a more important case) but for now it's sufficient to just make sure we don't use operators like toList() which risk leaks.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: Actually, if I throw in this scope, Netty will just log an error and the Single will never complete. So I might just make a top-level try-catch for everything inside the channel acquire callback.

RikkiGibson:
Some services will happily respond to us with a chunked encoding (no Content-Length), but not accept a request with a chunked encoding. So I think I had better not let our implementation rely on Content-Length..

RikkiGibson:
Also need to look into some test errors in autorest.java resulting from this

RikkiGibson:
Addressed #383

PR #387, opened 2018-02-22
Initial check in for resumable LRO.

alvadb:


azure-client-runtime/src/test/java/com/microsoft/azure/v2/AzureProxyTests.java
```diff
@@ -515,6 +526,76 @@ public void accept(OperationStatus<MockResource> operationStatus) {
...
+                        else {
+                            resource.set(operationStatus.result());
+                        }
+                    }
+                });
+
+        OperationDescription operationDescription = null;
+        PollStrategy.PollStrategyData pollData = null;
+        try {
+            operationDescription = mapper.readValue(data.toString(), OperationDescription.class);
```
RikkiGibson: Are you able to support a use case/add test coverage where an OperationDescription is assembled from user-provided values? One of the main users requesting this feature has asked that they be able to store the information received from a begin* method in whatever format they wish and convert it themselves to the format we need to resume polling.

anuchandy:
@alvadb @RikkiGibson - we might be already handling this, but just want to let you know as there was lot of confusion around it recently. If an LRO POST returns both async operation and location header, the progress will be tracked by doing polling on async operation link but at the end we are suppose to make a get on the location header to retrieve the result computed.

If you are looking for recently updated LRO doc from ARM, just talk to @shahabhijeet, he has been working with ARM recently for well documenting LRO.

azure-client-runtime/src/test/java/com/microsoft/azure/v2/AzureProxyTests.java
alvadb: A user would need to start a begin operation and then serialize the data to get the values for resuming the observing of the op.  The serialization to a JSON string would cover the case of manually supplying values as the JSON the user built up would need to be deserialized back to the OperationDescription object.

PR #393, opened 2018-03-02
Add RestResponse subtypes and handling

RikkiGibson:
This adds runtime support for concrete RestResponse subtypes.

The approach I'm using in the generator is basically thus:
- If a Headers type is defined, a Response type will be generated to contain the Headers and Body
- If no Headers type is defined, we will use `BodyResponse<TBody>`
- If only a `Flowable<ByteBuffer>` body is defined, we will use `StreamResponse`.
- If neither Headers or Body are defined we will use `VoidResponse`.

I've done a good amount of testing through autorest.java but will make sure to add some unit tests, perhaps updating RestProxyTests to use these new types before merging.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/BodyResponse.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/BodyResponse.java
@@ -15,14 +15,14 @@ import java.util.Map;
  */
 public final class BodyResponse<TBody> extends RestResponse<Void, TBody> {
     /**
-     * Create a new RestResponse object.
+     * Creates a new BodyResponse object.
      *
-     * @param statusCode The status code of the HTTP response.
-     * @param rawHeaders The raw headers of the HTTP response.
-     * @param tBody      The deserialized body.
+     * @param statusCode the status code of the HTTP response
+     * @param rawHeaders the raw headers of the HTTP response
+     * @param body the deserialized body
      */
-    public BodyResponse(int statusCode, Map<String, String> rawHeaders, TBody tBody) {
-        super(statusCode, null, rawHeaders, tBody);
+    public BodyResponse(int statusCode, Map<String, String> rawHeaders, TBody body) {
+        super(statusCode, null, rawHeaders, body);
     }
 
     // Used for uniform reflective creation in RestProxy.
@@ -32,8 +32,7 @@ public final class BodyResponse<TBody> extends RestResponse<Void, TBody> {
     }
 
     /**
-     * The deserialized body of the HTTP response.
-     * @return The deserialized body of the HTTP response.
+     * @return the deserialized body of the HTTP response
      */
     public TBody body() {
         return super.body();
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/BodyResponse.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/BodyResponse.java
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.rest.v2;
+
+import java.util.Map;
+
+/**
+ * A response to a REST call with a strongly-typed body specified.
+ *
+ * @param <TBody> The deserialized type of the response body.
+ */
+public final class BodyResponse<TBody> extends RestResponse<Void, TBody> {
+    /**
+     * Create a new RestResponse object.
+     *
+     * @param statusCode The status code of the HTTP response.
+     * @param rawHeaders The raw headers of the HTTP response.
+     * @param tBody      The deserialized body.
+     */
+    public BodyResponse(int statusCode, Map<String, String> rawHeaders, TBody tBody) {
+        super(statusCode, null, rawHeaders, tBody);
+    }
+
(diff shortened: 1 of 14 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/BodyResponse.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param <TBody> The deserialized type of the response body.
+ */
+public final class BodyResponse<TBody> extends RestResponse<Void, TBody> {
+    /**
+     * Create a new RestResponse object.
+     *
+     * @param statusCode The status code of the HTTP response.
+     * @param rawHeaders The raw headers of the HTTP response.
+     * @param tBody      The deserialized body.
```
RikkiGibson: should rename this to `body`

client-runtime/src/main/java/com/microsoft/rest/v2/RestResponse.java
```diff
@@ -39,63 +34,30 @@ public RestResponse(int statusCode, THeaders headers, Map<String, String> rawHea
     }
 
     /**
-     * The status code of the HTTP response.
      * @return The status code of the HTTP response.
```
RikkiGibson: Javadoc with repetition like this should be trimmed down. checkstyle requires the `@return` but not the description. Will figure out what the best practice is here.

client-runtime/src/main/java/com/microsoft/rest/v2/RestResponse.java
```diff
@@ -39,63 +34,30 @@ public RestResponse(int statusCode, THeaders headers, Map<String, String> rawHea
...
      * @return The deserialized body of the HTTP response.
      */
     public TBody body() {
         return body;
     }
-
-    /**
-     * Closes the content stream associated with this RestResponse, if any.
-     */
-    public void close() {
```
RikkiGibson: This is now handled only in concrete responses where we know the body is a `Flowable<ByteBuffer>`.

client-runtime/src/main/java/com/microsoft/rest/v2/StreamResponse.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public Void headers() {
+        return super.headers();
+    }
+
+    /**
+     * @return the body content stream
+     */
+    @Override
+    public Flowable<ByteBuffer> body() {
```
RikkiGibson: I'm adding these trivial overrides because I think it will provide a better editor experience if go-to-definition shows a concrete type instead of `TBody` or similar.

RikkiGibson:
As part of our preview release we're going to target Java 8 even though Joda time and some of the other changes have not yet been incorporated

client-runtime/src/main/java/com/microsoft/rest/v2/RestResponse.java
daschult: Anu mentioned this before. Many editors use the comment's description (the part you've removed) as the description in the Autocompletion/Intellisense popup. I believe the only place the `@return` piece is used is in Javadocs. I think it's a mistake to remove the description.

client-runtime/src/main/java/com/microsoft/rest/v2/protocol/HttpResponseDecoder.java
```diff
@@ -283,6 +285,7 @@ private Type getHeadersType() {
         }
 
         if (token.isSubtypeOf(RestResponse.class)) {
+            token = token.getSupertype(RestResponse.class);
```
daschult: It doesn't look like RestResponse has a supertype/base class. What does this return?

client-runtime/src/main/java/com/microsoft/rest/v2/BodyResponse.java
```diff
@@ -0,0 +1,40 @@
...
+package com.microsoft.rest.v2;
+
+import java.util.Map;
+
+/**
+ * A response to a REST call with a strongly-typed body specified.
+ *
+ * @param <TBody> The deserialized type of the response body.
+ */
+public final class BodyResponse<TBody> extends RestResponse<Void, TBody> {
```
daschult: Did you create GitHub issues for incorporating this into the generator?

client-runtime/src/main/java/com/microsoft/rest/v2/protocol/HttpResponseDecoder.java
RikkiGibson: In this case `token` is something like `BodyResponse` or `BlobGetResponse` and we're walking up the hierarchy to get the `RestResponse` superclass so that we can look at the type parameters in a uniform way.

client-runtime/src/main/java/com/microsoft/rest/v2/BodyResponse.java
RikkiGibson: https://github.com/Azure/autorest.java/issues/153

client-runtime/src/main/java/com/microsoft/rest/v2/RestResponse.java
RikkiGibson: OK, I think bringing it back for a better experience would be the right thing. I wonder which editors that's true for. In IntelliJ:
screen shot 2018-03-05 at 10 15 42 am

Issue #389, opened 2018-02-22
v2: NettyClient.ResponseContentFlowable no reentrancy protection


davidmoten:
Just had a look at `NettyClient.ResponseContentFlowable`. Implementations of `Flowable` are always tricky things so it drew my eye.

I noticed that it is marked as not thread safe but the reentrancy synchronous use case can provoke stack overflow at the moment.

In particular, `request(1)` calls `emitContent` which calls `subscriber.onNext` which may call `request(1)` and bingo after a few thousand items are emitted you have stack overflow. 

This is not hard to fix for the synchronous case. Instead of emitting immediately, add the `data.content()` to a queue and call a `drain()` method. If currently draining (check a boolean field) then do nothing otherwise emit the contents of the queue.

davidmoten:
I notice now that the only use of `ResponseContentFlowable` overrides the `request` and `cancel` methods (without referencing `super`) and does no request accounting.

RikkiGibson:
Thanks for the guidance. I've had problems with `request` calling `subscriber.onNext` in the past which I think I worked around with unnecessary use of `.observeOn()`. I'll have another look considering the approach you've suggested.

Regarding the request accounting: the indirection around request() and cancel() in here could be made more clear and maybe refactored to have this key information strewn about fewer areas, but if I'm understanding what you're saying correctly, this part looks sound to me.

- The ResponseContentFlowable constructor takes a Subscription as an argument which is used to interpret the Rx `request` and `cancel` methods in Netty terms (channel read, channel close). Since ResponseContentFlowable is also a Subscription it has implementations of `request` and `cancel` that do request accounting and call into the `request` and `cancel` on the given Subscription.

- ResponseContentFlowable contains `long chunksRequested` which is referenced in [channelReadComplete](https://github.com/Azure/autorest-clientruntime-for-java/blob/v2/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java#L488) after each Netty read completes to determine if another Netty read needs to be initiated

RikkiGibson:
Addressed in #406

PR #396, opened 2018-03-05
Various minor improvements

RikkiGibson:
- Gets rid of Apache Commons, which we appear to only have ever used in one instance in the entire runtime. Using Java 8's String.join() instead.
- Changed `Completable HttpClientFactory.shutdown()` to `void HttpClientFactory.close()` as this is not a context where optimizing the utilization of threads is useful and it's better to have something straightforward that conforms to the `Closeable` interface

### DIFF
--- a/client-runtime/pom.xml
+++ b/client-runtime/pom.xml
@@ -72,10 +72,6 @@
       <artifactId>jackson-dataformat-xml</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-lang3</artifactId>
-    </dependency>
-    <dependency>
       <groupId>io.reactivex.rxjava2</groupId>
       <artifactId>rxjava</artifactId>
     </dependency>
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpClientFactory.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpClientFactory.java
@@ -6,12 +6,12 @@
 
 package com.microsoft.rest.v2.http;
 
-import io.reactivex.Completable;
+import java.io.Closeable;
 
 /**
  * Creates an HttpClient from a Configuration.
  */
-public interface HttpClientFactory {
+public interface HttpClientFactory extends Closeable {
     /**
      * Creates an HttpClient with the given Configuration.
      * @param configuration the configuration.
@@ -20,11 +20,8 @@ public interface HttpClientFactory {
     HttpClient create(HttpClientConfiguration configuration);
 
     /**
-     * Asynchronously awaits completion of in-flight requests,
-     * then closes shared resources associated with this HttpClient.Factory.
-     * After this Completable completes, HttpClients created from this Factory can no longer be used.
-     *
-     * @return a Completable which shuts down the factory when subscribed to.
+     * Awaits completion of in-flight requests, then closes shared resources associated with this HttpClient.Factory.
+     * After this method returns, HttpClients created from this Factory can no longer be used.
      */
-    Completable shutdown();
+    void close();
 }
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -590,13 +590,8 @@ public final class NettyClient extends HttpClient {
         }
 
         @Override
-        public Completable shutdown() {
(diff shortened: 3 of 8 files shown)


RikkiGibson:
Also fixes a few issues based on feedback from @davidmoten

davidmoten:
Had a look at the current state of `ResponseContentFlowable` and you still don't have reentrancy protection and may have stall issues. What I was expecting is below. Firstly I assume that calls to `request` are serialized (as comments on class suggest). If request calls are not serialized then a different pattern is used (that I can provide).

The basic pattern is:

```java

long requested;
volatile boolean cancelled;
boolean draining;
final Queue<HttpContent> queue = new ArrayDeque<>();

public request(long n) {
    if (n <=0) return;
    requested+=n;
    drain();
}

void onReceivedContent(HttpContent data) {
  queue.add(data);
  drain();
}

void drain() {
  if (draining) return;
  draining = true;
  try {
   HttpContent content;
   while (!cancelled && requested > 0 && (content=queue.poll())!=null) {
      subscriber.onNext(content);
      requested--;
    }
  } finally {
    draining = false;
  }
}
```

RikkiGibson:
You're right--#389 hasn't been addressed yet.

The serialization is achieved through the fact that the Netty EventLoop runs all the Netty event handlers (e.g. `channelReadComplete`), and `.subscribeOn(Schedulers.from(ctx.eventLoop()))` causes `request()` to also run on the EventLoop. I could be misunderstanding this operator, though, or unwittingly using some bad practice.

I have a sense that this could be done better without requiring `request()` to run on the Netty EventLoop through `.subscribeOn()`. If you'd like to share or link to an example of the pattern for non-serialized queueing/draining (where a different thread calls `request()` than the thread that calls `onNext()`) it'd be greatly appreciated.

Perhaps a test that tries to write a large `Flowable.fromIterable()` to the channel would be useful to test the synchronous case?

davidmoten:
For interest's sake, the non-serialized pattern looks like below. However, I strongly suggest you use RxJava operators to do what you want to do because developing custom Flowables is a very risky business and you don't seem to be doing anything particularly unusual (from afar!). I doubt that you can't use existing operators. Check the factory method `generate`, `defer` and the `Subject`s are very powerful tools in conjunction with the rest of the library. If you can describe what you want to do in an abstract and condensed enough way I might be able to recommend an approach.

```java
AtomicLong requested = new AtomicLong();
volatile boolean cancelled;
final AtomicInteger wip = new AtomicInteger();
//use an Spsc queue if calls to `onReceivedContent` are serialized, otherwise use Mpsc queue
final SimplePlainQueue<HttpContent> queue = new SpscLinkedArrayQueue<>(); 

public request(long n) {
    if (n <=0) return; // actually to be reactive streams compliant you need to throw here I think
    BackpressureUtils.getAndAddRequest(requested, n);
    drain();
}

void onReceivedContent(HttpContent data) {
  queue.add(data);
  drain();
}

void drain() {
  if (wip.getAndIncrement() ==0) {
    int missed = 1;
    while (true) {
      long r = requested.get();
      long e;
      while (e != r) {
        if (cancelled) return;
        HttpContent o = queue.poll();
        if (o==null) break;
        subscriber.onNext(o);
        e++;
      }
      BackpressureUtils.produced(requested, e);
      int missed = wip.addAndGet(-missed);
      if (missed == 0) return;
    }
  }
}
```

RikkiGibson:
The main reason I've refrained from using the `generate` APIs is that it appears to be required to synchronously call one of the methods on the emitter when my method is called.

Something like a Subject would be handy here. Is there such a thing for Flowables? I'd like to simply be able to translate `request()` calls coming from the Subscription into `channel.read()` calls in Netty, and to call `onNext()` on the Subscriber asynchronously.

davidmoten:
Sounds like a job for `PublishSubject`. Note also that if the emissions to a subject are themselves non-serialized you can fix that with `subject.serialized()`.

In short, `flatMap/concatMap` (which buffers) and `PublishSubject` (which also buffers) are your friends. I've used `concatMap` because I assume you don't want out of order returns from the reads.

I haven't looked at what `channel.read()` does but you can adjust:

```java
  Flowable
    .rangeLong(0, Long.MAX_VALUE) 
    .concatMap(i -> {
         PublishSubject<String> subject = PublishSubject.create();
         channel.addListener(list -> {
             for (String s:list) {
                 subject.onNext(s);
             }
             subject.onComplete();
         });
         channel.read();
         return subject.toFlowable(BackpressureStrategy.BUFFER);
    });
```

PR #398, opened 2018-03-06
Update README.md and add convenience HttpPipelineBuilder methods

### LINKED ISSUE #397: Make standard RequestPolicies more discoverable?
If users were to try and build an HttpPipeline they would probably benefit from having all the standard RequestPolicies be easily discoverable. I suggest doing this by adding a class like this, similar to `Executors` or `Channels` classes in the JDK or `Functions` in RxJava:

```java
public final class RequestPolicies {
    private RequestPolicies() { }

    public static RequestPolicyFactory credentials(ServiceClientCredentials creds);
    public static RequestPolicyFactory cookies();
    public static RequestPolicyFactory decoding();
    public static RequestPolicyFactory httpLogging(HttpLogDetailLevel level);
    ...
}
```

Then the usage would look like:

```java
new HttpClientBuilder()
    .withRequestPolicy(RequestPolicies.credentials(creds))
    .withRequestPolicy(RequestPolicies.cookies())
    .withRequestPolicy(RequestPolicies.decoding())
    .withRequestPolicy(RequestPolicies.httpLogging(HttpLogDetailLevel.HEADERS))
    .build();
```

Thoughts?

RikkiGibson:
Fixes #397

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpPipelineBuilder.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpPipelineBuilder.java
@@ -6,11 +6,23 @@
 
 package com.microsoft.rest.v2.http;
 
+import com.microsoft.rest.v2.credentials.ServiceClientCredentials;
+import com.microsoft.rest.v2.policy.CookiePolicyFactory;
+import com.microsoft.rest.v2.policy.CredentialsPolicyFactory;
+import com.microsoft.rest.v2.policy.DecodingPolicyFactory;
+import com.microsoft.rest.v2.policy.HostPolicyFactory;
+import com.microsoft.rest.v2.policy.HttpLogDetailLevel;
+import com.microsoft.rest.v2.policy.HttpLoggingPolicyFactory;
+import com.microsoft.rest.v2.policy.ProxyAuthenticationPolicyFactory;
+import com.microsoft.rest.v2.policy.RequestIdPolicyFactory;
 import com.microsoft.rest.v2.policy.RequestPolicyFactory;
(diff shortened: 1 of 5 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/http/HttpPipelineBuilder.java
```diff
@@ -128,12 +140,110 @@ public HttpPipelineBuilder withRequestPolicies(RequestPolicyFactory... requestPo
...
+
+    /**
+     * Adds a RequestPolicy which fails a request if it does not complete by
+     * the time the given interval elapses.
+     * @param timeout The amount of time to wait before timing out a request.
+     * @param unit The unit of time associated with the timeout parameter.
+     * @return This HttpPipeline builder.
+     */
+    public HttpPipelineBuilder withTimeoutPolicy(long timeout, TimeUnit unit) {
+        return withRequestPolicy(new TimeoutPolicyFactory(timeout, unit));
```
daschult: Now that we're using Java 8, should this be changed to use Duration?

RikkiGibson: I guess the main reason for keeping it this way is alignment with the RxJava `.timeout()` API. If it were easy to translate the `Duration` to `(long, TimeUnit)` I'd be open to replacing or adding an overload.

daschult: It would probably be easy enough to convert the Duration to milliseconds, but let's leave it for now.

PR #402, opened 2018-03-08
Fix reentrant safety in ResponseContentFlowable

### LINKED ISSUE #389: v2: NettyClient.ResponseContentFlowable no reentrancy protection

RikkiGibson:
Fixes #389

- Converted the SingleOnSubscribe to lambda--sorry about the spurious diffs simply due to unindenting.
- Added a missing `subscription.cancel()` for the request content when writing to channel fails so that we stop reading from the request stream
- Introduced `draining` flag in NettyClient.java:402. This makes it so we don't push on additional stack frames for each item we emit from the queue in the case where the response `subscriber.onNext` makes a synchronous call to `subscription.request()`. The queue generally contains a small number of items, so this wasn't a big crash risk, but it's incorrect to leave it the way it originally was.

(Still interested in the potential for a solution that relies more heavily on RxJava operators for response content streaming, but it will have to wait until after beta 1).

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -38,6 +38,7 @@ import io.reactivex.Scheduler;
 import io.reactivex.Single;
 import io.reactivex.SingleEmitter;
 import io.reactivex.disposables.Disposable;
+import io.reactivex.internal.util.BackpressureHelper;
 import io.reactivex.schedulers.Schedulers;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -394,10 +395,7 @@ public final class NettyClient extends HttpClient {
             if (l <= 0) {
                 subscriber.onError(new IllegalArgumentException("Request amount must be positive. Received amount: " + l));
             } else {
-                chunksRequested += l;
-                if (chunksRequested < 0) {
-                    chunksRequested = Long.MAX_VALUE;
-                }
+                chunksRequested = BackpressureHelper.addCap(chunksRequested, l);
 
                 if (!draining) {
                     try {
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -37,9 +37,7 @@ import io.reactivex.FlowableSubscriber;
 import io.reactivex.Scheduler;
 import io.reactivex.Single;
 import io.reactivex.SingleEmitter;
-import io.reactivex.SingleOnSubscribe;
 import io.reactivex.disposables.Disposable;
-import io.reactivex.functions.Function;
 import io.reactivex.schedulers.Schedulers;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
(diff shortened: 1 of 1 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -413,13 +391,28 @@ public void request(long l) {
...
-            while (!queuedContent.isEmpty() && chunksRequested > 0 && !isCanceled) {
-                emitContent(queuedContent.remove());
-            }
+            if (l <= 0) {
+                subscriber.onError(new IllegalArgumentException("Request amount must be positive. Received amount: " + l));
+            } else {
+                chunksRequested += l;
+                if (chunksRequested < 0) {
+                    chunksRequested = Long.MAX_VALUE;
+                }
```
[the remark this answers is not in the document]
RikkiGibson: It might be more hygienic to replace with a call to RxJava's `BackpressureUtils.addCap()`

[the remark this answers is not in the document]
RikkiGibson: https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/util/BackpressureHelper.java#L28

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -413,13 +392,25 @@ public void request(long l) {
...
-            chunksRequested += l;
-            while (!queuedContent.isEmpty() && chunksRequested > 0 && !isCanceled) {
-                emitContent(queuedContent.remove());
-            }
+            if (l <= 0) {
+                subscriber.onError(new IllegalArgumentException("Request amount must be positive. Received amount: " + l));
+            } else {
+                chunksRequested = BackpressureHelper.addCap(chunksRequested, l);
+
+                if (!draining) {
```
davidmoten: This drain section should be in its own method and called by `request` and also called by `onReceivedContent` otherwise the stream can stall.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -413,13 +392,25 @@ public void request(long l) {
...
             }
 
-            chunksRequested += l;
-            while (!queuedContent.isEmpty() && chunksRequested > 0 && !isCanceled) {
-                emitContent(queuedContent.remove());
-            }
+            if (l <= 0) {
+                subscriber.onError(new IllegalArgumentException("Request amount must be positive. Received amount: " + l));
+            } else {
+                chunksRequested = BackpressureHelper.addCap(chunksRequested, l);
```
davidmoten: I don't see proper accounting happening here on the requests. This is a flatMap scenario where the number of items delivered from upstream is unknown so you make a request of 1 to upstream that puts n items on a queue. When the queue is emptied (having reported the items to downstream) and there are still outstanding requests from downstream you request 1 more from upstream. You can get fancy and request more just before the queue is exhausted but that's what RxJava operators will do for you (so don't bother, wait for rewrite).

davidmoten: I can make a PR to help with this if you like.

RikkiGibson: If you are willing to go to the trouble, it'd be greatly appreciated.

davidmoten: Sure, I'll have a few questions on the way.  I'll start with a read about Netty. Should be educational.

RikkiGibson: Feel free to ask on here or by email :)

davidmoten: I understand a bit more about what you are doing now. `ResponseContentFlowable` does need to have async protection as the `HttpResponse.streamBodyAsync` is public api and anything could happen to that exposed Flowable (it could be composed with any operator). I see you've got one subscription only protection which makes sense but `request` and `cancel` could come from any thread at all. Given that, it doesn't make sense that `request` have a check in it that the request has come from the channel eventLoop. Am I missing something here?

davidmoten: I see what you mean now about emitting `ByteBuf`s that need to be released and how in `NettyResponse` you map the `ByteBuf` items and release the `ByteBuf`s. This immediate release will have to stay I think but I'd probably move it into `ResponseContentFlowable`.

RikkiGibson: Re: the async protection. I have been always applying a `subscribeOn(eventLoop)` to the `ResponseContentFlowable` before returning it to a consumer which (if I understand the operator correctly) ensures that the `request()` calls from downstream will get dispatched to the event loop. I think there are more elegant ways of doing this-- I have been considering either just dispatching `request()` calls to the event loop if we're not currently on the event loop or (hopefully) we can just find a solution that obsoletes ResponseContentFlowable and the thread safety comes for free.

The `cancel()` calls I think can still come from any thread, so things like the final `release()` loop on any unemitted content are explicitly run on the event loop.

RikkiGibson: Re: emitting ByteBufs, it is definitely tricky to do properly. I have observed some memory leaks in heavy download scenarios involving transient cancellation, where I see that I'm enqueuing the ByteBuf in the `queuedContent` and somewhere down the line failing to release it. It would be nice to be able to stream the ByteBufs in private APIs of `NettyResponse`, because it allows us to reduce copying with common `bodyAsString()` and `bodyAsByteArray()` scenarios, but if this doesn't turn out to be tenable, we can live with it.

The original issue I posted RxJava is where I found that intermediate queues can drop these ByteBufs on cancellation--that includes the queue contained in `FlowableObserveOn` and others. I had found issues internally where a `Flowable.toList()` was dropping the "values accumulated so far" on cancellation. I think if we always apply the copy-and-release before exposing the stream in a public API, and take care about which operators are applied to it in internal APIs, it might be workable.

davidmoten: Unfortunately, applying a `subscribeOn(eventLoop)` has no effect on where requests come from (it just guarantees where subscription happens from). For example if downstream has an `observeOn` then requests can come from a non-eventLoop thread.

There are other ways to force requests to be on the event loop but won't be necessary as calling `request` just increases an `AtmoicLong` and `drain` is called where the necessary happens-before/eventLoop control can happen.

I've drafted up the operator and close to sorted, covering your concerns including releasing ByteBufs properly. I noticed the reuse of `Flowable<ByteBuf>` in `NettyResponse` in methods that  use reduced copying so it makes sense for the `Flowable` to be of type `ByteBuf`.  I'm going to launch into a lot of unit tests on it next. I'll probably submit in a few days (working on it in spare time).

>I think if we always apply the copy-and-release before exposing the stream in a public API, and take care about which operators are applied to it in internal APIs, it might be workable.

I agree, absolutely.

RikkiGibson: Makes sense, and much appreciated. You may also benefit from `RestProxyWithNettyTests` which are how we have been testing a wider variety of behaviors of the HTTP client in practice. The `NettyClientTests` actually came about only quite recently when I was testing shutdown behaviors.

PR #404, opened 2018-03-09
Remove Joda dependency

daschult:


azure-client-authentication/src/main/java/com/microsoft/azure/v2/credentials/AzureCliToken.java
```diff
@@ -72,11 +71,7 @@ String accessToken() {
 
     Date expiresOn() {
```
RikkiGibson: Weird. I guess this class won't be in our library when the new standalone auth library is released, at least.

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodParser.java
```diff
@@ -43,6 +46,7 @@
  * method.
  */
 public class SwaggerMethodParser {
+    private final SerializerAdapter<?> serializer;
```
RikkiGibson: Is this for handling template substitutions on non-String parameters?

azure-client-authentication/src/main/java/com/microsoft/azure/v2/credentials/AzureCliToken.java
daschult: Yeah. I left this one as Date on purpose.

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodParser.java
daschult: Correct. Specifically non-String header values and non-String query values.

client-runtime/src/main/java/com/microsoft/rest/v2/serializer/DateTimeSerializer.java
```diff
@@ -29,17 +29,33 @@
...
+     * Convert the provided OffsetDateTime to its String representation.
+     * @param offsetDateTime The OffsetDateTime to convert.
+     * @return The String representation of the provided offsetDateTime.
+     */
+    public static String toString(OffsetDateTime offsetDateTime) {
+        String result = null;
+        if (offsetDateTime != null) {
+            offsetDateTime = offsetDateTime.withOffsetSameInstant(ZoneOffset.UTC);
+            result = DateTimeFormatter.ISO_INSTANT.format(offsetDateTime);
+            if (result.startsWith("+")) {
```
RikkiGibson: Welp

RikkiGibson: Has anyone out in the wild reported/discussed/rationalized this behavior?

daschult: The ISO 8601 standard says that outside of 4 digit years any two parties using these serialized strings can determine whether they want to use + in front of their years. Joda and Autorest didn't use + signs, but it looks like the Java OffsetDateTime does use it. I'd rather not, so I just strip the leading + if it exists.

RikkiGibson:
Seems all good to me--too bad we had to roll own implementations of a couple more conversions.

Issue #405, opened 2018-03-09
NettyClient: Handling early response from server


RikkiGibson:
The server is allowed to send a response before we have finished sending the request. It looks like in the case where we receive an error response while sending a request body we should stop sending the request body. We might not be doing this correctly in NettyClient right now.

See https://stackoverflow.com/questions/14250991/is-it-acceptable-for-a-server-to-send-a-http-response-before-the-entire-request#14483857

I'm content to simply interpret 4xx and 5xx as being an error response instead of consulting the Swagger spec. We're at the HTTP layer after all, and not in RestProxy.

What's not clear to me is whether we need to close the connection in this scenario or if we can reuse it. I'm inclined to be cautious and simply close the connection, especially because we can't necessarily count on servers out in the wild being compliant with the dark corners of RFCs.

I have seen people mention scenarios where an HTTP server will actually give a "success" response before the request body is fully sent, because the server is transforming the body content on the fly and sending it back. Unless I can find 1. something that states "this is allowed in HTTP and should work" and 2. a service team that wants this, I'm inclined to not take any special effort to support it.

RikkiGibson:
We use the Disposable returned from the `Single<HttpResponse>.subscribe()` to cancel the request, so if we continue sending the request content even after the response was received, there's no way to cancel that request any more. So it's probably required to just always cancel the request and close the channel if a response was received before we finished sending the request.

RikkiGibson:
We're now closing the connection when we finish reading the response and the request has not yet been written. I think this will be adequate.

Issue #395, opened 2018-03-02
Fail fast when request content size doesn't match Content-Length


RikkiGibson:
When a request content stream doesn't match the expected Content-Length on the request, our runtime just hangs because the server is waiting to receive more content before giving a response. Let's add some counting/validation so that if a request content stream completes prematurely we will end the response subscription right away with some kind of exception.

rickle-msft:
I have observed this behavior in the case where the data is smaller than the specified size. I have also seen some odd behavior when the data is larger than the specified size. The PUT operation seems to have no problem. A subsequent download will mysteriously use the HTTP verb FAULTGET, which doesn't actually exist as far as I can tell. It was suggested this is because of some stream corruption that overflows into the next request.

RikkiGibson:
Resolved in #430

PR #406, opened 2018-03-16
v2 NettyClient flowable rework

davidmoten:
As promised in discussion in #402, this is a rework of `NettyClient.ResponseContentFlowable`.

The Flowable has been reworked to
* perform proper request accounting
* ensure reads are requested one at a time of upstream only on completion of last requested read (because `channel.read()` is ignored by Netty if a read is pending)
* buffer `ByteBuf`s as upstream delivers many items at a time (a *chunk*)
* release unused `ByteBuf`s on cancellation 
* handle asynchronous calls to `subscribeActual`, `request`, `cancel` with non-blocking techniques
* handle a connection failure by also propagating the `channelInactive` response to the Flowable (which emits an `IOException`)
* reject multiple subscriptions by emitting an error
* be compliant with reactive-streams contract (bad requests reported to RxJavaPlugins.onError)
* protect against events appearing unexpectedly after a final event (`done` flag)
* remove `eventLoop` usage because Netty does not require that for example `channel.read()` is called from the `eventLoop`.
* use *WireMock* in the unit tests to remove dependency on `httpbin.org`

Notes:
* `RestProxyWithNettyTests` are hanging because the `Future` returned by `channel.acquire` never completes (and the new `Flowable` does not get called at all). I'd like to hand that problem back to you as it appears to be independent of this change.
* There is a TODO in the `drain` method about whether we shortcut errors or not. Please read that.
* You may want to play with `ResourceLeakDetector`. Probably need a sleep at the end of tests to ensure that cancellation finishes.
* `NettyAdapter.sendRequestInternalAsync` is messy. I left it alone but for a couple of deprecation fixes and the reporting of `channelInactive`. I'd suggest a cleanup to reduce the method length and improve readability. If you go the extra mile to be friendly to users then you replace the anonymous classes (lambdas) with concrete classes to make stack traces easier to read.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -275,8 +275,6 @@ public final class NettyClient extends HttpClient {
                 if (transition(s, ACQUIRING_DISPOSED, CHANNEL_RELEASED)) {
                     channelPool.closeAndRelease(channel);
                     return;
-                } else if (transition(s, CHANNEL_RELEASED, CHANNEL_RELEASED)) {
-                    return;
                 } else if (transition(s, ACQUIRING_NOT_DISPOSED, ACQUIRED_CONTENT_NOT_SUBSCRIBED)) {
                     break;
                 } else if (state.compareAndSet(s,  s)) {
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -415,7 +415,7 @@ public final class NettyClient extends HttpClient {
                             break;
                         }
                     } else {
-                        break;
+                        throw new RuntimeException("unexpected!");
                     }
                 }
             }
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -299,77 +299,81 @@ public final class NettyClient extends HttpClient {
                 if (request.body() == null) {
                     writeBodyEnd();
                 } else {
-                    request.body().subscribe(createSubscriber(inboundHandler));
+                    request.body().subscribe(new RequestSubscriber(inboundHandler));
                 }
             } catch (Exception e) {
                 emitError(e);
             }
         }
+        
+        private final class RequestSubscriber implements FlowableSubscriber<ByteBuffer> {
+            Subscription subscription;
+            
+            // we need a done flag because an onNext emission can throw and emit an Error 
+            // event though the onNext cancels the subscription that is best endeavours for the 
+            // upstream so we need to be defensive about terminal events that follow
+            private boolean done;
 
-        private FlowableSubscriber<ByteBuffer> createSubscriber(final HttpClientInboundHandler inboundHandler) {
-            return new FlowableSubscriber<ByteBuffer>() {
-                Subscription subscription;
-                
-                // we need a done flag because an onNext emission can throw and emit an Error 
-                // event though the onNext cancels the subscription that is best endeavours for the 
-                // upstream so we need to be defensive about terminal events that follow
-                private boolean done;
-                
-                @Override
-                public void onSubscribe(Subscription s) {
-                    subscription = s;
-                    inboundHandler.requestContentSubscription = subscription;
-                    subscription.request(1);
-                }
+            private final HttpClientInboundHandler inboundHandler;
+            
+            public RequestSubscriber(HttpClientInboundHandler inboundHandler) {
+                this.inboundHandler = inboundHandler;
+            }
 
-                GenericFutureListener<Future<Void>> onChannelWriteComplete = (Future<Void> future) -> {
-                    if (!future.isSuccess()) {
-                        subscription.cancel();
-                        emitError(future.cause());
-                    }
-                };
+            @Override
+            public void onSubscribe(Subscription s) {
+                subscription = s;
+                inboundHandler.requestContentSubscription = subscription;
+                subscription.request(1);
+            }
 
-                @Override
-                public void onNext(ByteBuffer buf) {
-                    if (done) {
-                        return;
-                    }
-                    try {
-                        channel
-                            .writeAndFlush(Unpooled.wrappedBuffer(buf))
-                            .addListener(onChannelWriteComplete);
-
-                        if (channel.isWritable()) {
-                            subscription.request(1);
-                        }
-                    } catch (Exception e) {
-                        subscription.cancel();
-                        onError(e);
-                    }
-                }
-
-                @Override
-                public void onError(Throwable t) {
-                    if (done) {
-                        return;
-                    }
-                    done = true;
-                    emitError(t);
-                }
-
-                @Override
-                public void onComplete() {
-                    if (done) {
-                        return;
-                    }
-                    done = true;
-                    try {
-                        writeBodyEnd();
-                    } catch (Exception e) {
-                        emitError(e);
-                    }
+            GenericFutureListener<Future<Void>> onChannelWriteComplete = (Future<Void> future) -> {
+                if (!future.isSuccess()) {
+                    subscription.cancel();
+                    emitError(future.cause());
                 }
             };
+
+            @Override
+            public void onNext(ByteBuffer buf) {
+                if (done) {
+                    return;
+                }
+                try {
+                    channel
+                        .writeAndFlush(Unpooled.wrappedBuffer(buf))
+                        .addListener(onChannelWriteComplete);
+
+                    if (channel.isWritable()) {
+                        subscription.request(1);
+                    }
+                } catch (Exception e) {
+                    subscription.cancel();
+                    onError(e);
+                }
+            }
+
+            @Override
+            public void onError(Throwable t) {
+                if (done) {
+                    return;
+                }
+                done = true;
+                emitError(t);
+            }
+
+            @Override
+            public void onComplete() {
+                if (done) {
+                    return;
+                }
+                done = true;
+                try {
+                    writeBodyEnd();
+                } catch (Exception e) {
+                    emitError(e);
+                }
+            }
         }
 
         private void writeRequest(final DefaultHttpRequest raw) {
@@ -547,6 +551,7 @@ public final class NettyClient extends HttpClient {
         }
         return raw;
     }
+    
 
     /**
      * Emits HTTP response content from Netty.
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -70,7 +70,7 @@ public final class NettyClient extends HttpClient {
 
     /**
      * Creates NettyClient.
-     * 
+     *
      * @param configuration
      *            the HTTP client configuration.
      * @param adapter
@@ -209,8 +209,8 @@ public final class NettyClient extends HttpClient {
             });
         }
     }
-    
-    private static  void addHeaders(final HttpRequest request) {
+
+    private static void addHeaders(final HttpRequest request) {
         request.withHeader(io.netty.handler.codec.http.HttpHeaderNames.HOST.toString(), request.url().getHost())
                .withHeader(io.netty.handler.codec.http.HttpHeaderNames.CONNECTION.toString(),
                         io.netty.handler.codec.http.HttpHeaderValues.KEEP_ALIVE.toString());
@@ -237,11 +237,11 @@ public final class NettyClient extends HttpClient {
         private final SharedChannelPool channelPool;
         private final HttpRequest request;
         private final SingleEmitter<HttpResponse> responseEmitter;
-        
-        // state is tracked to ensure that any races between write, read, 
-        // disposal, cancel, and request are properly handled via a serialized state machine. 
+
+        // state is tracked to ensure that any races between write, read,
+        // disposal, cancel, and request are properly handled via a serialized state machine.
         private final AtomicInteger state = new AtomicInteger(ACQUIRING_NOT_DISPOSED);
-        
+
         private static final int ACQUIRING_NOT_DISPOSED = 0;
         private static final int ACQUIRING_DISPOSED = 1;
         private static final int ACQUIRED_CONTENT_NOT_SUBSCRIBED = 2;
@@ -252,7 +252,7 @@ public final class NettyClient extends HttpClient {
 
         // synchronized by `state`
         private Channel channel;
-        
+
         //synchronized by `state`
         private ResponseContentFlowable content;
         
@@ -294,12 +294,12 @@ public final class NettyClient extends HttpClient {
                     return;
                 }
             }
-            
-            final HttpClientInboundHandler inboundHandler = 
+
+            final HttpClientInboundHandler inboundHandler =
                     channel.pipeline().get(HttpClientInboundHandler.class);
             inboundHandler.setFields(responseEmitter, this);
             //TODO do we need a memory barrier here to ensure vis of responseEmitter in other threads?
-            
+
             try {
                 configurePipeline(channel, request);
 
@@ -320,9 +320,9 @@ public final class NettyClient extends HttpClient {
         
         private final class RequestSubscriber implements FlowableSubscriber<ByteBuffer>, GenericFutureListener<Future<Void>> {
             Subscription subscription;
-            
-            // we need a done flag because an onNext emission can throw and emit an Error 
-            // event though the onNext cancels the subscription that is best endeavours for the 
+
+            // we need a done flag because an onNext emission can throw and emit an Error
+            // event though the onNext cancels the subscription that is best endeavours for the
             // upstream so we need to be defensive about terminal events that follow
             private boolean done;
 
@@ -333,7 +333,7 @@ public final class NettyClient extends HttpClient {
             private static final int WRITING_WRITABLE = 1;
             private static final int WRITE_COMPLETED_NOT_WRITABLE = 2;
             private static final int WRITING_NOT_WRITABLE = 3;
-            
+
             RequestSubscriber(HttpClientInboundHandler inboundHandler) {
                 this.inboundHandler = inboundHandler;
             }
@@ -363,7 +363,7 @@ public final class NettyClient extends HttpClient {
 
             @Override
             public void onError(Throwable t) {
-                // TODO should we wrap the throwable so that the client 
+                // TODO should we wrap the throwable so that the client
                 // knows that the error occurred from the request body?
                 if (done) {
                     RxJavaPlugins.onError(t);
@@ -520,11 +520,11 @@ public final class NettyClient extends HttpClient {
                 }
             }
         }
-        
+
         /**
          * Returns false if and only if content subscription should be immediately
          * cancelled.
-         * 
+         *
          * @param content
          * @return false if and only if content subscription should be immediately
          *     cancelled
@@ -547,10 +547,10 @@ public final class NettyClient extends HttpClient {
                 }
             }
         }
-        
+
         /**
          * Is called when content flowable terminates or is cancelled.
-         * 
+         *
          **/
         void contentDone() {
             while (true) {
@@ -601,7 +601,7 @@ public final class NettyClient extends HttpClient {
                 }
             }
         }
-        
+
         private void closeAndReleaseChannel() {
             channelPool.closeAndRelease(channel);
         }
@@ -644,7 +644,7 @@ public final class NettyClient extends HttpClient {
         }
         return raw;
     }
-    
+
 
     /**
      * Emits HTTP response content from Netty.
@@ -688,8 +688,8 @@ public final class NettyClient extends HttpClient {
                 subscriber = s;
                 s.onSubscribe(this);
 
-                acquisitionListener.contentSubscribed(this); 
-                
+                acquisitionListener.contentSubscribed(this);
+
                 // now that subscriber has been set enable the drain loop
                 wip.lazySet(0);
 
@@ -772,14 +772,14 @@ public final class NettyClient extends HttpClient {
 
         private void drain() {
             // Below is a non-blocking technique to ensure serialization (in-order
-            // processing) of the block inside the if statement and also to ensure 
+            // processing) of the block inside the if statement and also to ensure
             // no race conditions exist where items on the queue would be missed.
             //
             // wip = `work in progress` and follows a naming convention in RxJava
             //
-            // `missed` is a clever little trick to ensure that we only do as many 
+            // `missed` is a clever little trick to ensure that we only do as many
             // loops as actually required. If `drain` is called say 10 times while
-            // the `drain` loop is active then we notice that there are possibly 
+            // the `drain` loop is active then we notice that there are possibly
             // extra items on the queue that arrived just after we found none left
             // (and before the method exits). We don't need to loop around ten times
             // but only once because all items will be picked up from the queue in
@@ -829,11 +829,11 @@ public final class NettyClient extends HttpClient {
                         }
                     }
                     if (e > 0) {
-                        // it's tempting to use the result of this method to avoid 
-                        // another volatile read of requested but to avoid race conditions 
+                        // it's tempting to use the result of this method to avoid
+                        // another volatile read of requested but to avoid race conditions
                         // it's essential that requested is read again AFTER wip is changed.
                         BackpressureHelper.produced(requested, e);
-                    } 
+                    }
                     missed = wip.addAndGet(-missed);
                     if (missed == 0) {
                         return;
@@ -870,17 +870,17 @@ public final class NettyClient extends HttpClient {
             }
         }
     }
-    
+
     private static final class ChannelSubscription implements Subscription {
-        
+
         private final AtomicReference<Channel> channel;
         private final AcquisitionListener acquisitionListener;
-        
+
         ChannelSubscription(AtomicReference<Channel> channel, AcquisitionListener acquisitionListener) {
             this.channel = channel;
             this.acquisitionListener = acquisitionListener;
         }
-        
+
         @Override
         public void request(long n) {
             Preconditions.checkArgument(n == 1, "requests must be one at a time!");
@@ -906,10 +906,10 @@ public final class NettyClient extends HttpClient {
         private volatile Subscription requestContentSubscription;
 
         private AtomicReference<Channel> channel = new AtomicReference<Channel>();
-        
+
         HttpClientInboundHandler() {
         }
-        
+
         void setFields(SingleEmitter<HttpResponse> responseEmitter, AcquisitionListener acquisitionListener) {
             // this will be called before request has been initiated
             this.responseEmitter = responseEmitter;
@@ -931,7 +931,13 @@ public final class NettyClient extends HttpClient {
 
         @Override
         public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
-            contentEmitter.chunkCompleted();
+            if (contentEmitter != null) {
+                // It doesn't seem like this should be possible since we set this volatile field
+                // before we begin writing request content, but it can happen under high load
+                contentEmitter.chunkCompleted();
+            } else {
+                LoggerFactory.getLogger(getClass()).warn("contentEmitter was null!");
+            }
         }
 
         @Override
@@ -949,7 +955,7 @@ public final class NettyClient extends HttpClient {
                     exceptionCaught(ctx, response.decoderResult().cause());
                     return;
                 }
-                // Scheduler scheduler = Schedulers.from(ctx.channel().eventLoop());
+
                 responseEmitter.onSuccess(new NettyResponse(response, contentEmitter));
             }
 
@@ -962,7 +968,7 @@ public final class NettyClient extends HttpClient {
 
             if (msg instanceof LastHttpContent) {
                 acquisitionListener.contentDone();
-            } 
+            }
         }
 
         @Override
@@ -991,7 +997,7 @@ public final class NettyClient extends HttpClient {
         /**
          * Create a Netty client factory, specifying the event loop group size and the
          * channel pool size.
-         * 
+         *
          * @param eventLoopGroupSize
          *            the number of event loop executors
          * @param channelPoolSize
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -17,6 +17,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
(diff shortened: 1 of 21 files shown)


RikkiGibson:
Thanks for the remarkable level of attention to detail here. I'll be reviewing in detail in the near future.

Agreed on `sendRequestInternalAsync`. The GenericFutureListener should perhaps be divided out into a few additional methods or classes.

ResourceLeakDetector is very good-- I still have some leaks detected by it showing up in stress tests that I need to iron out.

Looks like CI has been stalling out when testing this, which I'll also have to track down. Maybe related to WireMock?

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -353,98 +360,206 @@ public void onComplete() {
...
+                            subscriber.onError(error);
+                            return;
+                        }
+                        HttpContent o = queue.poll();
+                        if (o != null) {
+                            e++;
+                            if (emitContent(o)) {
+                                return;
+                            }
+                        } else {
```
RikkiGibson: It looks like the drain loop continues even when the queue is empty (`poll()` returns null). Is that the right thing to do?

This seems to be causing a hanging issue where `responseEmitter.onSuccess()` synchronously subscribes to the `ResponseContentFlowable` and requests the max amount. Then, this drain loop runs forever because the Netty thread never frees up to service the `channelRead` and put content into the queue.

RikkiGibson:
@davidmoten Would be useful to know if `mvn test -pl client-runtime -Dtest=NettyClientTests` passes in your environment. It's hanging in mine.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
davidmoten: Spot on. There needs to be a `break` there.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -353,98 +360,206 @@ public void onComplete() {
...
-            });
+                int missed = 1;
+                long r = requested.get();
+                while (true) {
+                    long e = 0;
+                    while (e != r) {
+                        // Note that an error can shortcut the emission of content that is currently on the queue.
+                        // This is probably desirable generally because it prevents work that being done downstream 
+                        // that might be thrown away anyway due to the error. 
+                        // TODO discuss with project members
```
RikkiGibson: I think the behavior you've implemented is the preferred one-- if an error occurred in the producer, release the queued ByteBufs instead of propagating them, and propagate the error instead.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: Excellent-- my tests look all good locally with this latest change.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
davidmoten: good oh, I'll remove the todo line

davidmoten:
thanks @RikkiGibson, I've
* added the missing `break` to the drain loop (oops!)
* added `awaitDone` to a couple of unit tests
* removed TODO line about discussing delaying errors

Now everything passes including `RestProxyWithNettyTests` so I was probably wrong about it being unrelated! I stuffed up debugging I guess. I ran the NettyClientTests in a loop for half an hour and is ok.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -353,98 +360,206 @@ public void onComplete() {
...
 
-            currentEventLoop.execute(() -> {
-                while (!queuedContent.isEmpty()) {
-                    queuedContent.remove().release();
+        private void requestChunkOfByteBufsFromUpstream() {
+            channelSubscription.request(1);
+        }
+        
+        private void drain() {
+            if (wip.getAndIncrement() == 0) {
```
RikkiGibson: Can you help me understand what `wip` and `missed` means?

RikkiGibson:
Only remaining concern think I have is that tests like `testFlowableResponseShortBodyAsByteArrayAsync` seem to take 2+ seconds to complete in my environment. Any idea what that could be about?

davidmoten:
I've added comments to the drain method about `wip` and `missed`. Here's a copy
```java
private void drain() {
    // Below is a non-blocking technique to ensure serialization (in-order processing) of the block inside the if statement
    // and also to ensure no race conditions exist where items on the queue would be missed.
    //
    // wip = `work in progress` and follows a naming convention in RxJava
    //
    // `missed` is a clever little trick to ensure that we only do as many loops as actually required. If `drain` is called 
    // say 10 times while the `drain` loop is active then we notice that there are possibly extra items on the queue that arrived
    // just after we found none left (and before the method exits). We don't need to loop around ten times but only once because
    // all items will be picked up from the queue in one additional polling loop.
```
So the effect of `missed = wip.addAndGet(-missed)` is to convert those 10 outstanding drain calls into one extra loop around.

davidmoten:
>Only remaining concern think I have is that tests like testFlowableResponseShortBodyAsByteArrayAsync seem to take 2+ seconds to complete in my environment. Any idea what that could be about?

I noticed those tests were a bit slow. Something to do with `WireMock` perhaps? I'll be offline (out bushwalking) for a couple of days so won't be able to do investigations but can continue chat for another couple of hours while I get ready.

RikkiGibson:
I think we can live with another 15~ seconds of test runtime in the short term while I track it down. Looks like the remaining CI issues are just due to Checkstyle stuff (lack of spaces around != operator and such). I'm happy to fix those up myself and merge as soon as I've verified the behavior is good when running RestProxyStressTests in a VM.

RikkiGibson:
Actually, thinking on it, I've found that frequent creation/destruction of a NettyClient.Factory can be time consuming, so I will probably try that to get the test execution time down. If no customization or lifecycle management of the NettyClient.Factory is necessary it's good to use `HttpClient.createDefault()`.

davidmoten:
checkstyle fixed

davidmoten:
I'll modify to use `HttpClient.createDefault`.

davidmoten:
Yep that brought the time down by 2 secs per test.

RikkiGibson:
Great. Thanks so much for taking the time to do all the little revisions.

I'm currently seeing hanging issues when running `RestProxyStressTests.download100MParallelTest` on a 50~ MB/s connection. I would be willing to triage this a little. I can also try and modify that test suite to use WireMock and see if the issue is still reproducible that way.

There's no rush, so we can just follow up on this when you're back. If I decide to do some work on this I may simply merge into a feature branch in the Azure repo and open another PR from there.

Thanks again for your help and enjoy your trip!

davidmoten:
More issues that come to mind:
* should the body stream emit an error if the number of bytes emitted is not equal to the value in the Content-Length header?
* I haven't checked if the `channelUnregistered` event has meaning for us
* I also shy away from using the word `async` in a public api method that returns a `Flowable`. One of the beauties of returning a `Flowable` is that you can make sync/async behaviour an implementation detail.

RikkiGibson:
Will see if I can take some time today to track down those test issues, particularly trying to repro with WireMock.

re Content-Length: It might help debugging bad server behaviors to put such a check on the response-body side of things. I definitely want this for requests (see #395) as it's currently a bit easy for people to inadvertently provide a bad Content-Length and the current behavior is for Netty to just hang. If this is implemented it's important to just do nothing in the `Transfer-Encoding: chunked` case. :)

re channelUnregistered: this event appears to be in response to the channel no longer being associated with the EventLoop and underlying Selector. We never migrate channels between EventLoops ourselves so I don't think there's anything meaningful we can do here.

re async: Does "sync behavior" mean it occupies the thread that calls `.subscribe()` or that is occupies some thread on a pool somewhere? Or either of these?

RikkiGibson:
I was able to gain a little traction with WireMock for stress testing, but when I try to use the "file response body" feature it runs out of memory when serving downloads at meaningful levels of parallelism (e.g. 10+). I think I was able to repro the hanging issue when running the test serially, but would like to go for something that can scale a little higher that feels a little less messy.

I'm going to attempt to use maven-jetty-plugin with a simple jetty.xml to serve up the contents of a folder under a particular web server path. I'll let you know how it goes.

davidmoten:
>re Content-Length: It might help debugging bad server behaviors to put such a check on the response-body side of things. I definitely want this for requests (see #395) as it's currently a bit easy for people to inadvertently provide a bad Content-Length and the current behavior is for Netty to just hang. If this is implemented it's important to just do nothing in the Transfer-Encoding: chunked case. :)

Yeah I reckon emitting an error is too strict. Clients can do their own check.

>re channelUnregistered: this event appears to be in response to the channel no longer being associated with the EventLoop and underlying Selector. We never migrate channels between EventLoops ourselves so I don't think there's anything meaningful we can do here.

Ah good-oh, ta.

>re async: Does "sync behavior" mean it occupies the thread that calls .subscribe() or that it occupies some thread on a pool somewhere? Or either of these?

An example of a synchonronous Flowable is `Flowable.range()` which emits on the subscribing thread. `Flowable.interval(10, TimeUnit.SECONDS)` actually emits on a thread from `Schedulers.computation()` but it wasn't called `Flowable.intervalAsync`. 

I expect `NettyResponse.streamBodyAsync` will end up emitting on the Netty EventLoop as it stands which means async but an implementation detail is that you might cache a small response completely ahead of time and emit it on the subscribing thread (which means the Flowable is synchronous). The point is the client doesn't need to know that it emits async or sync (and should not assume either) when they use a `Flowable` and the name should not constrain the implementation.

Speaking of emitting on the Netty EventLoop. If you always wanted to handover the output from the EventLoop to another thread you would place an `observeOn` in `NettyResponse`. You may want to do this because I believe that Netty mandates that EventLoop threads should not be blocked and if you emit on the EventLoop the client may block downstream (e.g. write the received bytes across a network). I'd be tempted to put an `observeOn` in `NettyClient.streamBodyAsync` or else explain in the javadoc of `NettyClient.streamBody` that downstream actions should not block the emitting thread.

Anyway, I'd suggest that this method:
```java
public abstract Flowable<ByteBuffer> streamBodyAsync();
```
should be called 
```java
public abstract Flowable<ByteBuffer> body();
```

RikkiGibson:
The "Async" suffix is a .NET-ism that leaked into our Java codebase a long time ago--I'm happy to be rid of it in this case.

In some other cases, like in a client library, I'm concerned about how people will interpret the semantics. Consider a method like `Single<VirtualMachine> VirtualMachines.get(String resourceId)` which has no effects until you subscribe. (We currently "reserve" this name for synchronous overloads of these methods, which I'm also pushing to do away with.)

I'm about to push up a commit to this branch in my fork which should allow you to exercise some higher-intensity download behaviors locally. There's probably an approach which doesn't require you to write so many files to your system, but I wanted to test how things behave under a lot of disk load as well as a lot of network load.

davidmoten:
> In some other cases, like in a client library, I'm concerned about how people will interpret the semantics. Consider a method like Single<VirtualMachine> VirtualMachines.get(String resourceId) which has no effects until you subscribe. (We currently "reserve" this name for synchronous overloads of these methods, which I'm also pushing to do away with.)

Yep I can understand that the semantics might be a tad troublesome. However:
* when a user tries to use a Flowable `body()` they will go hey this is not a string/byte array what do I do with it (and read the docs, learn about RxJava, etc)
* be helpful in the javadoc and the overall documentation and they will sort it out
* loading the names up with hints just adds a lot of noise and confuscation for the user's code

This semantic confusion is part and parcel of reactive programming. I vote for less noise but your are welcome to differ!

RikkiGibson:
I think you're right, particularly that proper Javadocs will make it clear enough. Hopefully we will be able to change our approach to overloads to permit a simpler naming convention in generated clients (Azure/autorest.java#129)

davidmoten:
I've added a concurrency test to NettyClientTests called `testConcurrentRequests` that uses WireMock's jetty server. It pushes through 1GB in 10s across 10 threads. You can scale up the number of requests as much as you want for a long running test. 

This might be handy as a template for parallel work with RxJava2 as well.

davidmoten:
Hmm, long running test runs out of memory. Either WireMock or us. I'll investigate.

davidmoten:
Was WireMock, needed to disable request journalling.

davidmoten:
Now asserting md5 hashes, ran 100GB through in 733s. Memory use fine.

davidmoten:
renamed NettyResponse.streamBodyAsync to NettyResponse.body and added javadoc to explain gotchas. Happy to roll this change back if a bit early to consider renames.

RikkiGibson:
That's excellent-- I'm good with the rename, I'll try and review the recent changes pretty soon here. Note that the type that client library users are likely to see at this point looks more like [StreamResponse.java](https://github.com/Azure/autorest-clientruntime-for-java/blob/v2/client-runtime/src/main/java/com/microsoft/rest/v2/StreamResponse.java). Mainly people will encounter this HttpResponse type when they're handling errors (it's contained inside the RestException object we produce) or when they're implementing an interceptor that reads body content.

I'm starting to get curious if there's a way to really put this through its paces by creating a more "antagonistic" consumer for testing which requests on multiple threads at the same time or what have you.

RikkiGibson:
Had a look at the recent commits. Looks good. Just a checkstyle violation was introduced.

davidmoten:
I'm fixing race conditions in `sendRequestInternalAsync` and have this question:

If a response completes it is possible that the request has not completed writing. Do we need to allow for this situation (and not close the channel till writing finished)?

RikkiGibson:
No, we don't have any services that require support for this. I suggest that if the request is not finished writing at the time the response completes, let's just close the channel. I wrote a bit about this problem in #405.

davidmoten:
Ok, many changes here. I've modified `NettyClient` heaps more. I've extracted methods and made concrete classes to improve readability of code and stacktraces and added a state machine to `sendRequestInternalAsync` with `compareAndSet` transitions to deal with a bunch of race conditions. 

So got` NettyClientTest`s passing and *same* behaviour from download stress test! I dug deeper as mentioned above into Netty itself and saw 2 race conditions where the `SingleThreadedEventExecutor` can stall. I submitted https://github.com/netty/netty/pull/7815 to fix these but I suspect there are more because I haven't seen the sort of race protection that RxJava uses (wip+missed).

The stress tests pass when I add a `channel.read()` at the end of `HttpClientInboundHandler.channelRead()` (which has a TODO for removal). This invalidates backpressure of course so the read buffers into the queue and the client is in trouble if it processes slowly and is dealing with a really big read (enough to blow the available heap).

There suspect there are more concurrency bugs in Netty itself that are blocking our backpressure attempts. I'll keep looking. 

So this PR is not ready for merge still (and I will review it more myself next week) but is getting there!

RikkiGibson:
Thanks for the update--looking forward to reviewing :)

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -204,13 +202,19 @@ private NettyAdapter(int eventLoopGroupSize, int channelPoolSize) {
             }).onErrorResumeNext((Throwable throwable) -> {
                 if (throwable instanceof EncoderException) {
                     LoggerFactory.getLogger(getClass()).warn("Got EncoderException: " + throwable.getMessage());
+                    //TODO what is this, a retry? Should have a time delay? Max number of retries?
```
RikkiGibson: I've been getting EncoderException under load in cases where the server has responded with 503 or otherwise failed to receive the request. As far as I can tell it relates to the internal state of the Netty HttpRequestEncoder getting corrupted. I've been resolving it by throwing away the connection/pipeline and grabbing a new one.

There are probably some conditions (i.e. the early response from server case that seems to keep coming up) where  we should just be closing the channel and discarding the related HttpRequestEncoder but we're failing to do so.

davidmoten: Weird eh. As long as you are happy we don't end up in some high cpu infinite loop because we don't inject a time delay...

davidmoten:
Ok found it. After much reading of Netty source code I came to the conclusion no problem there (and my PR was no problem either because I wasn't aware of the full lifecycle). Injected lots of printlns and realized that the race condition was of my own creation. 

Simple fix. It was essential that I don't use the results of `BackpressureHelper.produced` to calculate the next value of `r` because to avoid race conditions `requested` must be read *after* `wip` is changed.

Ran the download test for 2 hours on my laptop without problem.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
RikkiGibson: I'd like most of all to root out what's putting the encoder into a bad state when the server sends an error response early, and then remove the usage of `onErrorResumeNext` here. If it can't be solved in a reasonable amount of time I will just add a delay on each retry. Thanks for pointing this one out.

RikkiGibson:
That's great. Sometime soon I'll try and pull latest, get on a cloud VM and run the other stress tests against a storage account. That'll help ensure there's no additional complications related to transient faults or in scenarios like writing a download stream directly to an upload stream.

It would be great to get the local tests to parity with the Azure-dependent tests. Ideally WireMock or something similar could receive a large request stream and pass a Content-MD5 for it in the response. If it's also possible to make WireMock sometimes respond early with an error or close the connection then we'd pretty much be there.

The CI failure is just a rename in client-runtime that didn't get extended to azure-client-runtime.

RikkiGibson:
I like what you've done extracting some of the anonymous classes out and naming them. Download test is looking great on the VM.

I'm seeing the following stack trace when running `RestProxyStressTests.upload100MParallelTest()`:
```
java.lang.NullPointerException
	at com.microsoft.rest.v2.http.NettyClient$HttpClientInboundHandler.channelReadComplete(NettyClient.java:838)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:398)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:380)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelReadComplete(AbstractChannelHandlerContext.java:373)
	at io.netty.handler.codec.ByteToMessageDecoder.channelReadComplete(ByteToMessageDecoder.java:324)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:398)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:380)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelReadComplete(AbstractChannelHandlerContext.java:373)
	at io.netty.handler.ssl.SslHandler.channelReadComplete(SslHandler.java:1217)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:398)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:380)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelReadComplete(AbstractChannelHandlerContext.java:373)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelReadComplete(DefaultChannelPipeline.java:1419)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:398)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelReadComplete(AbstractChannelHandlerContext.java:380)
	at io.netty.channel.DefaultChannelPipeline.fireChannelReadComplete(DefaultChannelPipeline.java:951)
	at io.netty.channel.epoll.AbstractEpollStreamChannel$EpollStreamUnsafe.epollInReady(AbstractEpollStreamChannel.java:820)
	at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:404)
	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:304)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:886)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
```

I'm going to try and do a little additional work to make the rest of the stress test suite run against the WireMock server, or the Jetty server if WireMock won't do the trick.

davidmoten:
Just had a quick look. Might need a null check on `contentEmitter` but will be good to understand when `channelInactive` gets called. I also notice that `FileReadFlowable` needs work (I think you've mentioned that before). It will need the standard `drain` loop with `wip` protection.

RikkiGibson:
Looked at WireMock's ResponseTransformer--doesn't appear to handle streaming request bodies so I don't think it will support the kind of upload behavior that's needed. Working on a little servlet to run on the local Jetty server that just sends back an MD5 of the response content--will also try and make it fail randomly as well.

On my local machine I'm seeing some interesting behavior related to uploads. `FlowableUtil.readFile()` seems to be working great for uploads in the synthetic upload100MParallelTest. However, uploadMemoryMappedTest is producing bad MD5s. `FlowableUtil.split(FileChannel.map())` is getting some kind of corruption due to the FlowableUtil.split implementation (which uses pretty simple Flowable.generate APIs).

Removing FlowableUtil.split() fixes the corruption, but it's necessary to use something like that for HTTPS so that the Netty SSL handler doesn't attempt to encrypt the entire request body all in one chunk, which makes us run out of memory. Also, limiting the upload rate with `stream.zip(Flowable.interval, (a, b) -> a)` seems to fix the corruption as well--which makes me think something racy is going on.

Re: FileReadFlowable - I'm optimistic it will be simpler to implement properly because it doesn't need to queue anything (at least the way it's written now). I don't see any reentrant safety issues in there, but I could be missing something. It would be much appreciated if you had a look.

RikkiGibson:
It would be great if Flowable.generate could be used to implement it, but it looks like it's limited to synchronous use cases.

Originally I had implemented FlowableUtil.readFile sort of like this:
```java
AsynchronousFileChannel fileChannel = ...
Flowable.range(0, NUM_CHUNKS)
    .flatMap(i -> Completable.fromFuture(
        fileChannel.read(ByteBuffer.allocate(CHUNK_SIZE), i * CHUNK_SIZE));
```

but this seems to block threads anyway, and I found this was considerably slower than a Flowable.generate that reads from a regular FileChannel.

davidmoten:
Hey @RikkiGibson, something to add to the TODO list is that `HttpResponse.bodyAsString()` is buggy.  You can't chop up bytes into chunks, convert the chunks into Strings and append the strings. Character sets are fiddlier than that. You can use [rxjava2-extras](https://github.com/davidmoten/rxjava2-extras) for that with 


```java
Flowable<byte[]> bytes;
Flowable<String> string = bytes.compose(
        x -> Transformers.decode(x, StandardCharsets.UTF_8));
```

RikkiGibson:
Which impl converts chunks to string and concatenates? It looks in NettyResponse like we append all the ByteBufs from the response content together and then call ByteBuf.toString(Charset).

davidmoten:
> Which impl converts chunks to string and concatenates? It looks in NettyResponse like we append all the ByteBufs from the response content together and then call ByteBuf.toString(Charset).

Oops you are right, overlooked `collectContent`. Now change the TODO so that a `Flowable<String>` is returned rather than `Single<String>` because otherwise we can blow out memory with large content. Perhaps an additional method?

davidmoten:
I've rewritten `FlowableUtil.readFile` and added tests. Please take it for a spin.

RikkiGibson:
That's interesting. So does your Transformers.decode() method do something like look for reasonable delimiters in the input, so that you can produce n number of Strings without reading the whole response content into memory?

RikkiGibson:
Thanks for taking a look at the FileReadFlowable--it should be useful and maybe later on can even be split out as part of an RxJava File I/O package.

Streaming body content as String could be worth incorporating down the line.

I'm sending you another PR davidmoten/autorest-clientruntime-for-java#2 which should allow for mocking of all the other stress testing methods. I'd like to figure out a relatively elegant way of running it. A little bit later I'm hoping to add some transient faults, but first I'd like to fix the happy paths.

Using the mock server on my Mac:
- download100MParallelTest looks good
- upload100MParallelTest passes when I take the FileReadFlowable from a4689c9 and fails (MD5 mismatch) when I take the one from fa2e325.
- uploadMemoryMappedTest fails due to an MD5 mismatch-- this makes me think the new impl of FileReadFlowable is probably fine, but there's something racy in the way we upload the request content.
- downloadUploadStreamingTest fails due to an MD5 mismatch

davidmoten:
In NettyClient.RequestSubscriber requests only happen if channel is writable so if the channel is not writable we might end up with a stall. The next request should be in the future listener for the write I assume. I did that and still see stalls so might be something else, I'll keep looking but my weekend is about to start so might have to wait a couple of days.

RikkiGibson:
For some reason, moving the `if (channel.isWritable()) subscription.request(1);` logic to the `onChannelWriteComplete` fixed the MD5 mismatch error. I find that a little strange, I guess, but I'm willing to roll with it as the perf still seems good.

Will play with transient faults now and see where I get.

RikkiGibson:
- It looks like a request body which ends with an error doesn't cause the responseEmitter to propagate the error through the responseEmitter. From the debugger it seemed like the emitError method is called while the AcquisitionListener is in state `ACQUIRED_DISPOSED_CONTENT_NOT_SUBSCRIBED`.
- Still tracking down some hangs related to uploads.

davidmoten:
>It looks like a request body which ends with an error doesn't cause the responseEmitter to propagate the error. From the debugger it seemed like the emitError method is called while the AcquisitionListener is in state ACQUIRED_DISPOSED_CONTENT_NOT_SUBSCRIBED.

Righto. We should just add a unit test that causes this. I think there are a number of cases still to be covered properly with unit testing. 

I'll look about for the MD5 mismatch error. Again a small unit test would be fab. Concurrency takes a while to get right but we'll get there.

RikkiGibson:
Ah, looks like you added the test coverage for the error in the request body. I've had my nose down in this work for a bit, otherwise I might have tried to share a PR for just that, since I also wrote a test for it.

It looks like with some very minor changes to what you have, mainly waiting to request more until after the channel write completes, the stress tests are all good on the Azure VM in Windows using the local MockServer. I'll probably try a few live tests against Azure Storage, then move on to Linux and try to get a handle on if there are any lingering issues on that side.

We are definitely getting close.

davidmoten:
> It looks like a request body which ends with an error doesn't cause the responseEmitter to propagate the error. From the debugger it seemed like the emitError method is called while the AcquisitionListener is in state ACQUIRED_DISPOSED_CONTENT_NOT_SUBSCRIBED.

Added two unit tests (that failed, the commit history doesn't show that because I copied and pasted from another branch!) where the request body emits an error straight away and the request body emits an error after some bytes.

The disposable was being set twice which immediately disposes the first disposable. Fixed this.

Noticed another problem that was a pretty stupid implementation on my part of the state transition where looping didn't happen on CAS fail. Fixed that too.

davidmoten:
@RikkiGibson I've added a `buffer.position(bytesWanted)` just before `buffer.flip()`. Can you check that for me please? This would be a source of extra bytes.

RikkiGibson:
I think I can see how the old behavior could result in unwanted additional bytes when only a subsection of the file is being read. Thanks for noticing that.

RikkiGibson:
I've incorporated your latest stuff. MD5 corruption seems to be under control. Mac and Windows are both passing all their stress tests.

- Downloads in Linux look good
- I seem to have stopped getting errors/log statements from contentEmitter being null in channelReadComplete-- did you do something that might've affected this and I overlooked it?
- I still get hangs in upload100MParallelTest on Linux, both live and local

I'm gonna call it a day for now but since Mac and Windows have both been tamed under pretty high load (1 GB/s or more throughput, with transient server-side failures, and no memory leaks under PARANOID leak detection), I think it's getting quite close.

RikkiGibson:
I sent you PR davidmoten/autorest-clientruntime-for-java#3 just so you can see what I've tweaked--besides the slightly obnoxious automatic trimming of whitespace, I've made an env var for the temp folder path. On a VM there tends to be a separate volume for fast temporary storage, and I think I was probably changing this path on the client but not the server, causing them each to grab the files and MD5s from different locations.. oops.

Notably, I added a .isWritable() check before requesting more request content in `onChannelWriteComplete`. It seems like this would be best to prevent overwhelming the channel. With or without this I am getting hangs in upload100MParallelTest on Linux.

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -172,306 +187,665 @@ private NettyAdapter(int eventLoopGroupSize, int channelPoolSize) {
...
 
-                        for (HttpHeader header : request.headers()) {
-                            raw.headers().add(header.name(), header.value());
+            GenericFutureListener<Future<Void>> onChannelWriteComplete = (Future<Void> future) -> {
+                if (!future.isSuccess()) {
+                    done = true;
+                    subscription.cancel();
+                    emitError(future.cause());
+                }
+                else if (channel.isWritable()) {
```
davidmoten: Trouble is this could stall the read of the request body. If we put this in then we need to detect the channel write status change from true to false and then request if the latest write has completed. A bit of coordination between onChannelWriteComplete and writability state change that might need a drain approach.

davidmoten: I noticed that we do request when channel writability changes but can over-request there too. I'll sort this out. Just need a CAS loop in AcquisitionListener.

davidmoten:
added writability+write completion atomic state machine to only request of upstream once channel is writable and last write complete

client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -379,6 +385,85 @@ public void onComplete() {
...
+                            if (writing.compareAndSet(s, WRITING_WRITABLE)) {
+                                break;
+                            }
+                        } else {
+                            break;
+                        }
+                    } else {
+                        if (s == WRITE_COMPLETED_WRITABLE) {
+                            if (writing.compareAndSet(s, WRITE_COMPLETED_NOT_WRITABLE)) {
+                                subscription.request(1);
```
RikkiGibson: We reach this line if the channel transitioned from writable to not writable and request content is not currently being written to the channel, right? So does it make sense to request more in that case?

RikkiGibson: Oops, I apparently was commenting on an outdated commit anyway.

davidmoten: well spotted, you were right and yep already fixed in 67fcb17

RikkiGibson:
Some observations related to the Linux hangs:
- I still get a decent number of "contentEmitter was null!" logs at scale for both downloads and uploads, which still baffles me a bit, but it doesn't actually seem to impact correctness
- I should've been using a beefier VM-- looks like I was getting throttled pretty badly after uploading 5~ GB. I would actually finish eventually even though after a while I was only uploading at 1Mb/s or so. After scaling up the VM, the live tests seem to be passing more promptly/reliably.
- When I use Netty's nio transport, all the tests pass on Linux. It seems that the hangs are only reproducible when using the native epoll transport. Commenting out the `client-runtime-native` test dependency in `client-runtime`, combined with the previous point, gets the tests passing reliably on Linux. I have seen some issues in Netty's GitHub repo related to deadlocks with the epoll transport, so I updated netty to latest, but it doesn't seem to have changed the behavior. It would be nice for us to run with the epoll transport, as it seems to really improve the performance, but if it turns out to just not be workable, that might be OK.

When I run upload100MParallelTest with the concurrency limited to 1 and start logging in `channel.writeAndFlush` and `GenericFutureListener.operationComplete`, I notice that the last log statement tends to be `channel.writeAndFlush`--as if for some reason one of the writes is simply not completing/flushing.

RikkiGibson:
Because I'm able to get all the stress tests passing on all the platforms in a reasonable configuration, I'm going to merge this now. Thanks a lot for your work and guidance @davidmoten. It's been invaluable.

I've got a few minor changes that I'm just going to PR separately related to connection pooling when performing large numbers of small requests.

PR #416, opened 2018-04-05
Add back apache commons for now

RikkiGibson:
I removed this dependency a little while back, but didn't actually stamp out its usages in the generated code, so I'm bringing it back until I can put in the time to remove it properly. This fixes build errors in some of our user's projects related to base64 encoding.

### DIFF
--- a/client-runtime/pom.xml
+++ b/client-runtime/pom.xml
@@ -72,6 +72,10 @@
       <artifactId>jackson-dataformat-xml</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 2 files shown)


daschult:
Oops! I accidentally merged it. I hope you don't mind.

RikkiGibson:
It's fine, but I think I'm missing a dependency that was actually needed:
```xml
<dependency>
  <groupId>commons-codec</groupId>
  <artifactId>commons-codec</artifactId>
  <version>1.10</version>
</dependency>
```
I remember working with @rickle-msft on this and we found that this was only specified in autorest.java, but not in the runtime, even though the generated code depends on it. Will need to revisit it soon.

PR #417, opened 2018-04-05
Move to Java 7 with threetenbp

RikkiGibson:
- Using threetenbp for datetime
- Removing compiler plugin config in child modules that seems redundant
- ~~Using retrolambda instead of rewriting all the lambdas~~
  - this seems to be working fine, but note that it won't run during a `mvn compile`. It will run during a `mvn test`, `mvn install` or `mvn package` though
  - We ignore IntelliJ IDEA stuff in git for now, but you can make sure you don't use Java 8 API accidentally by searching for the IntelliJ setting: "Usages of API which isn't available at the configured language level". Set it to "Higher than: 7 - Diamonds, ARM, multi-catch, etc."


RikkiGibson:
I started to see this kind of output when building autorest.java with retrolambda, so I think I'm rolling the retrolambda thing back.
screen shot 2018-04-05 at 3 33 31 pm

PR #420, opened 2018-04-12
Netty client rework 2

RikkiGibson:
Essentially a continuation of #406 

- Adds `finishedWritingRequestBody` flag to determine whether a channel should be closed when a response body is finished reading. This lets us pass `RestProxyStressTests.testHighParallelism` in a reasonable amount of time (~5 seconds) instead of either running for about a minute or just failing due to being unable to open more ports.
  - This perhaps should be changed so that we just close the connection if the request hasn't been fully written at the time we receive the response status/headers.
- Other minor refinements/trimming dead code
- Attempting to make RestProxyStressTests easier to run automatically, with everything able to run against the local test server.
  - Trying to get the test server to spawn automatically with the tests, but it's introducing new hanging issues, so it's probably a stretch goal for another PR.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -247,7 +247,8 @@ public final class NettyClient extends HttpClient {
 
         //synchronized by `state`
         private ResponseContentFlowable content;
-        
+
+        private volatile boolean finishedWritingRequestBody;
         private volatile RequestSubscriber requestSubscriber;
 
         AcquisitionListener(SharedChannelPool channelPool, final HttpRequest request,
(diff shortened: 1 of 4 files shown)


client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
```diff
@@ -554,6 +540,14 @@ boolean contentSubscribed(ResponseContentFlowable content) {
             }
         }
 
+        private void releaseAndCloseIfRequestNotComplete() {
+            if (finishedWritingRequestBody) {
+                channelPool.release(channel);
+            } else {
+                closeAndReleaseChannel();
```
RikkiGibson: There might be a race here where we close a channel that didn't necessarily need to be closed. finishedWritingRequestBody is assigned by one thread between this thread executing lines 544 and 547.

RikkiGibson:
There are more possible refinements/fixes here but will merge for now so we can publish another beta.

PR #425, opened 2018-04-21
Various minor improvements

RikkiGibson:
- NettyClient calls channel.write() until isWritable returns false, then flushes. Before, we flushed every single write. It appeared to be necessary to prevent data corruption. Now it seems like dispatching to the event loop before writing to the channel fixes the problem (this happens internally when you channel.write off of the event loop anyway).
  - [x] I still need to test this in Linux and Windows VM environments
- Removes client-runtime-native package. I wanted this to be a thing so it would be easier to get people access to the native modules, but I think the behavior of os-maven-plugin makes it hard to include in external projects. It's probably better that we just point out the available native modules directly to people.
  - [x] Need to update README.md to reflect this
- Base64 stuff is now consolidated to use java.util.Base64. I added a Base64Util class because I don't want to have to change all the usages if we need to migrate to yet another Base64 library.
- Minor goodies for @rickle-msft, fulfilling requests that were made months ago for HTTP date header (through a RequestPolicy) and HttpPipeline.build(Iterable).


client-runtime/src/main/java/com/microsoft/rest/v2/http/BufferedHttpResponse.java
```diff
@@ -49,43 +47,40 @@ public HttpHeaders headers() {
...
         if (body == null) {
             body = innerHttpResponse.bodyAsByteArray()
-                    .map(new Function<byte[], byte[]>() {
-                        @Override
-                        public byte[] apply(byte[] bytes) {
-                            body = Single.just(bytes);
-                            return bytes;
-                        }
+                    .map(bytes -> {
+                        body = Single.just(bytes);
```
daschult: I know you're not really changing this code, but just out of curiosity, isn't this a race condition? If the map is being run by a different thread, then it's possible that body will get assigned on line 49 after it gets assigned on line 51?

PR #430, opened 2018-05-21
Add CountingOperator and tests

### LINKED ISSUE #395: Fail fast when request content size doesn't match Content-Length

RikkiGibson:
Based on a request from Storage that cases where we accidentally provide the wrong Content-Length should result in an error, not in a hang or in the HTTP request completing without sending all the bytes in the stream.

Fixes #395

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
@@ -12,6 +12,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.reactivex.Completable;
 import io.reactivex.Flowable;
+import io.reactivex.FlowableOperator;
 import io.reactivex.FlowableSubscriber;
 import io.reactivex.Single;
 import io.reactivex.internal.subscriptions.SubscriptionHelper;
@@ -27,6 +28,7 @@ import java.nio.channels.AsynchronousFileChannel;
 import java.nio.channels.CompletionHandler;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Contains helper methods for dealing with Flowables.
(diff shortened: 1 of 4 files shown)


RikkiGibson:
@rickle-msft

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
```diff
@@ -70,6 +72,70 @@ public static boolean isFlowableByteBuffer(TypeToken entityTypeToken) {
...
+
+                @Override
+                public void onError(Throwable t) {
+                    observer.onError(t);
+                }
+
+                @Override
+                public void onComplete() {
+                    long count = bytesRead.get();
+                    if (count != bytesExpected) {
```
daschult: Should you be explicit here and do a `<` check instead of `!=`?

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
```diff
@@ -70,6 +72,70 @@ public static boolean isFlowableByteBuffer(TypeToken entityTypeToken) {
 
         });
     }
+    private static final class CountingOperator implements FlowableOperator<ByteBuffer, ByteBuffer> {
```
davidmoten: A tad strange to see `FlowableOperator` used. None of the operators are `lift` based in RxJava 2.x, adds small overhead and harder to think about. Standard is to use `compose`.

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
```diff
@@ -70,6 +72,70 @@ public static boolean isFlowableByteBuffer(TypeToken entityTypeToken) {
...
+                throw new IllegalArgumentException("bytesExpected must be non-negative, but was " + bytesExpected);
+            }
+
+            this.bytesExpected = bytesExpected;
+        }
+
+        @Override
+        public Subscriber<? super ByteBuffer> apply(Subscriber<? super ByteBuffer> observer) {
+            final AtomicLong bytesRead = new AtomicLong();
+            final AtomicReference<Subscription> subscription = new AtomicReference<>();
```
davidmoten: Put `subscription` as a non-volatile field in `Subscriber` because is part of the contract on upstream publisher that there is a happens-before relationship between `onSubscribe` and emissions.

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
```diff
@@ -70,6 +72,70 @@ public static boolean isFlowableByteBuffer(TypeToken entityTypeToken) {
...
+            if (bytesExpected < 0) {
+                throw new IllegalArgumentException("bytesExpected must be non-negative, but was " + bytesExpected);
+            }
+
+            this.bytesExpected = bytesExpected;
+        }
+
+        @Override
+        public Subscriber<? super ByteBuffer> apply(Subscriber<? super ByteBuffer> observer) {
+            final AtomicLong bytesRead = new AtomicLong();
```
davidmoten: Put as non-volatile field in `Subscriber`

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
RikkiGibson: Needs to reject on both the "too few" case and the "too many" case

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
RikkiGibson: Ah, thanks. I'll have a look and correct this.

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
RikkiGibson: `long` or `AtomicLong`? Is there a happens-before guarantee around subsequent calls to onNext? (and is there documentation somewhere I should know about that discusses the guarantees around the Flowable lifecycle, besides Reactive-Streams spec?)

davidmoten: absolutely, all signals (onSubscribe, onNext, onError, onComplete) are serialized individually and mutually. This is fundamental to reactive-streams. This shows signal serialization to you https://github.com/reactive-streams/reactive-streams-jvm#api-components. The spec is a tiny bit vague I think on this fundamental concept. I might raise an issue. The Observable Contract also covers it http://reactivex.io/documentation/contract.html.

davidmoten: From Observable Contract link:

>Observables must issue notifications to observers serially (not in parallel). They may issue these notifications from different threads, but there must be a formal happens-before relationship between the notifications.

RikkiGibson: That's really helpful to know. I thought point 11 of https://github.com/reactive-streams/reactive-streams-jvm#2-subscriber-code might also imply this but wasn't sure. Had also heard in the past that you can expect multiple threads will not call onNext concurrently--but didn't feel certain of whether this means a happens-before in the JMM sense.

davidmoten: Yeah, that point 11 is a bit of a distraction. It's about what a `Subscriber` does with arrived signals not about how they arrive. 1.3 covers what we want but I think the wording is imprecise.

RikkiGibson:
Is the nicest way actually to make ensureLength() return a Function which we pass right to .compose()?

client-runtime/src/main/java/com/microsoft/rest/v2/util/FlowableUtil.java
```diff
@@ -70,6 +72,73 @@ public static boolean isFlowableByteBuffer(TypeToken entityTypeToken) {
...
+
+                @Override
+                public void onSubscribe(Subscription s) {
+                    subscription = s;
+                    observer.onSubscribe(s);
+                }
+
+                @Override
+                public void onNext(ByteBuffer byteBuffer) {
+                    long newBytesRead = bytesRead + byteBuffer.remaining();
```
davidmoten: Because onNext can call onError you need protection here. The reason is if onNext routes to onError and the next event is say onComplete it is against the contract to emit events after the terminal event. What you do to prevent it is:

* declare non-volatile field `boolean done;`
* in onNext:
```java
if (done) {
  return;
}
...
```
* in onError:
```java
if (done) {
  RxJavaPlugins.onError(t);
  return;
}
done = true;
...
```
* in onComplete:
```java
if (done) {
  return;
}
done = true;
...
```

davidmoten:
By the way by using a custom operator you miss out on fusion which means you miss performance benefits. The best way might have been to use `defer` to wrap the state of `bytesExpected` and then the `doOnEach` operator inside.

RikkiGibson:
Good idea--the code is much simpler that way.

PR #439, opened 2018-06-13
Add additionalInfo to CloudError

j5lim:
'additionalInfo', a new property added to ARM error contract, contains additional error information. For example, it will contain detailed policy violation error information when the request to ARM failed due to violation of Azure policies.

The changes are to allow deserializing additionalInfo in ARM error based on the error type.

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/CloudError.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/CloudError.java
@@ -9,6 +9,9 @@ package com.microsoft.azure;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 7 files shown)


j5lim:
@jianghaolu @RikkiGibson Could anyone please review the PR?

RikkiGibson:
@jianghaolu feel free to add any comments after-the-fact and we can make changes later if there turns out to be a need.

jianghaolu:
Sorry I missed this PR. I don't think this should be merged. Asir was involved in some conversation about this. But it's already released in 1.6.2. I realized it when I was writing the release notes.

I will start a thread tomorrow to talk about this.

PR #450, opened 2018-07-25
Support handling threads by OkHttp

jianghaolu:


RikkiGibson:
Does this make it so threads:connections is no longer a 1:1 ratio?

jianghaolu:
It's still 1:1 - limiting number of connections indirectly through thread pool size.

The PR is to allow using that thread pool in OkHttpClient. Currently the threading is done by RxJava schedulers but people who are doing large-scale sync or simple parallel async can benefit from a hands-off management of threading.

jianghaolu:
OkHttp is always blocking IO in that matter.

PR #456, opened 2018-08-01
Update Netty and break apart large buffers

RikkiGibson:
- Update Netty version
- Use .split() to ensure we don't attempt to encode an excessively large buffer with SSL before sending
  - This magic number 8192 can be adjusted-- maybe 1024*64 is better?


RikkiGibson:
What do you think @rickle-msft

rickle-msft:
It's been working well for us!

PR #463, opened 2018-08-03
Issue fixes

jianghaolu:


azure-client-runtime/src/main/java/com/microsoft/azure/AzureEnvironment.java
```diff
@@ -97,6 +101,8 @@ public AzureEnvironment(Map<String, String> endpoints) {
         // TODO: add dns suffixes for the US government for datalake store and datalake analytics once they are defined.
         put("azureDataLakeStoreFileSystemEndpointSuffix", "N/A");
         put("azureDataLakeAnalyticsCatalogAndJobEndpointSuffix", "N/A");
+        put("azureLogAnalyticsResourceId", "N/A");
```
alexeldeib: https://api.loganalytics.us/

Need to double check if we have made App Insights available in US gov.

alexeldeib: Not available yet; just Log Analaytics to add here.

alexeldeib:
Looks good to me besides adding US gov 👍 Thanks!

PR #468, opened 2018-08-16
Add UnexpectedLengthException

RikkiGibson:
@rickle-msft

This enables the consumer of the ensureLength error to consume and rethrow with a better contextual message (e.g. in the context of retries)


RikkiGibson:
@rickle-msft I deployed a snapshot of this

PR #474, opened 2018-08-20
Enable MSICrdentials in v2: Merging msi credentials for app-service and vm

anuchandy:
Address the workitem - https://github.com/Azure/autorest-clientruntime-for-java/issues/471


azure-client-authentication/src/main/java/com/microsoft/azure/v2/credentials/MSICredentials.java
```diff
@@ -18,41 +16,145 @@
...
  * Managed Service Identity token based credentials for use with a REST Service Client.
  */
 @Beta
-public class MSICredentials extends AzureTokenCredentials {
-    private final String resource;
-    private final int msiPort;
-    private final AzureJacksonAdapter adapter;
+public final class MSICredentials extends AzureTokenCredentials {
+    //
+    private final List<Integer> retrySlots = new ArrayList<>(Arrays.asList(new Integer[] {1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597, 2584, 4181, 6765}));
```
jianghaolu: Wanted to ask when you were OOF - why Fibonacci here?

jianghaolu: Last release we made a patch to allow setting max retry - but max retry can go out of bound in this array. An exponential retry can avoid that and also a default max retry other than 6700 seconds may be a little more user-friendly

anuchandy: Agreed, let's switch to exponential and will take care of slot array OOB as well.

PR #475, opened 2018-08-21
Moving beta annotation to azure-client-runtime

anuchandy:


azure-client-authentication/src/main/java/com/microsoft/azure/v2/credentials/AuthFile.java
```diff
@@ -26,7 +25,7 @@
 /**
  * This class describes the information from a .azureauth file.
  */
-@Beta(SinceVersion.V1_1_0)
+@Beta("SinceVersion.V1_1_0")
```
RikkiGibson: good call.

RikkiGibson:
Anything blocking merge on this?

jianghaolu:
Anu is adding a few fields to the annotations. like

```java
String value();

String reason();
```

etc

Issue #451, opened 2018-07-31
Connecting with HTTPS on ports other than 443


RikkiGibson:
We make an assumption that port 443 means https while any other port means http.

https://github.com/Azure/autorest-clientruntime-for-java/blob/40a9487485a4487d95681350cbf11c1120e6de0b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java#L218

This should be changed to first check the protocol and if necessary fall back to port conventions.

RikkiGibson:
We added isProxyHTTPS in #462. Will keep an eye on whether this resolves the situation.

rickle-msft:
According to the customer, this issue was not resolved. Sercan sent out information on how we can repro this on a Linux vm

PR #476, opened 2018-08-22
Fix HTTPS proxying

RikkiGibson:
Related to #451 

I've finally gotten a grip on [how HTTPs proxying actually works](https://github.com/netty/netty/blob/4.1/handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java#L155) and am making changes as appropriate--including just relying on Netty's HttpProxyHandler to do the job.

This change seems like the right direction--we pass the destination address to channel.connect(), and pass the proxy address to HttpProxyHandler.

I've fixed the remaining issues by always using `HttpClientCodec` instead using of `HttpRequestEncoder` and `HttpResponseDecoder`. I think we were putting the pipeline in an inconsistent state by manipulating it from outside the event loop thread.


RikkiGibson:
I found that some tests started passing when I started adding handlers in an EventLoop task instead of in the worker thread for the SharedChannelPool. Very strange.

PR #479, opened 2018-08-24
Add Context class, HttpRequest property, and RestProxy handling

RikkiGibson:


RikkiGibson:
@rickle-msft please review

client-runtime/src/main/java/com/microsoft/rest/v2/SwaggerMethodParser.java
```diff
@@ -312,6 +312,20 @@ public String path(Object[] methodArguments) {
         return result;
     }
 
+    /**
+     * Get the {@link Context} passed into the proxy method.
+     * @param swaggerMethodArguments the arguments passed to the proxy method
+     * @return the Context, or null if no Context was provided
+     */
+    public Context context(Object[] swaggerMethodArguments) {
+        Object firstArg = swaggerMethodArguments[0];
```
RikkiGibson: This might break empty argument lists actually..

PR #482, opened 2018-08-28
Remove channel writing state machine

RikkiGibson:
This state machine doesn't appear to do anything because all the writes occur on the event loop anyway. We can rely on channel.isWritable() and the nature of the single event loop thread running for the channel to prevent multiple simultaneous requests for request content.

@rickle-msft

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,6 @@ sudo: false
 env:
 - TERM=dumb # Makes Gradle use plain console output
 script:
-- mvn clean install
-- mvn checkstyle:check
+- mvn clean install -B -Dstyle.color=always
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -11,6 +11,6 @@ sudo: false
 env:
 - TERM=dumb # Makes Gradle use plain console output
 script:
-- mvn clean install
-- mvn checkstyle:check
+- mvn clean install -B -Dstyle.color=always
```
RikkiGibson: Looks like the -B reduces the number of lines in Travis output from 15k to 3k. If I'm understanding right. It seems to mitigate the problem where Travis has to spin for a while before it can show the build log.

PR #504, opened 2018-10-23
Added configurability of SSL Context

gknapowski:
This pull request is coming form Isobar  The modifications contained will enable us to utilize this latest (v2) version of the Azure Storage on the AF CCE project.    The actual  changes are simple and they enable passing in an SslContext to the underlying NettyClient implementation of HttpClient.  This was done via an overloaded constructor.  Existing constructor signatures were left and reference the new implementation by passing in null for the SslContext which will cause the SSL behavior to fallback to the previous (now default) behavior.

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/v2/http/NettyClient.java
@@ -26,6 +26,8 @@ import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpVersion;
 import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.util.concurrent.DefaultThreadFactory;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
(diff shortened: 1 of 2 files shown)


jianghaolu:
+ @anuchandy do you think this is something that should be added in `ChannelPoolOptions`?

jianghaolu:
I think the changes look fine - and we will likely merge this - but there's a change we'd like to do in light of this -

we already have a class https://github.com/Azure/autorest-clientruntime-for-java/blob/v2/client-runtime/src/main/java/com/microsoft/rest/v2/http/SharedChannelPoolOptions.java where we put all the configurations we need there. However, it's not exposed to public. We plan to move SSLContext setting to that class as well.

gknapowski:
When I created my patch, the SharedChannelPoolOptions did not exist so I had to merge my code with it.  If it had been public I would have done exactly as you are saying and made the SslContext setable on it.

anuchandy:
@jianghaolu - Given we want to make ChannelPoolOptions public, should we put sslcontext into a HttpOptions class and make HttpOptions to compse channelpooloptions type? Was thinking that pool options should be specific to pool configuration and http related stuff (such as connection time out, readtimeout, ssl context ) should be put in httpOptions. That way we have a clear separation between two stuff, Http & Pool. What do you think?

anuchandy:
@gknapowski we're happy to take your PR. This is a great addition to the library/runtime. We want to make sure when we release the bits, library expose it in an extensible way (e.g via HttpOptions/ChannelPoolOptions configuration), that way we can avoid any breaking changes going forward. So we are thinking of merging your PR then Jianghao and I can work together to move your contribution to configuration object. You can review it then. Let us know if this is fine with you.

gknapowski:
That's great, I think using a command pattern is the best way to do this.  The sooner the better for our project as there are a lot of eyes on it including at Microsoft.

anuchandy:
@gknapowski could you retarget this PR to the branch https://github.com/Azure/autorest-clientruntime-for-java/tree/sslcontext-netty ? Once it is merged, we will work on configuration related stuff in this branch.

anuchandy:
@gknapowski thank you.

Issue #552, opened 2019-02-18
[v3] Can HttpMethod type be an enum


anuchandy:
#### [Details from @JonathanGiles]:

1. Why is HttpMethod not an enum? (I quickly rewrote it as such and expected compilation errors, but they never occurred). 
2. This also means that ExpandableStringEnum could also be removed.

anuchandy:
@JonathanGiles sure it can be enum but it's being ExpandableStringEnum allow us to accommodate new values. Fine with enum or ex-enum.

`ExpandableStringEnum` is an important type  to support expandable enum types in swagger spec. There are data plane and mgmt plane swaggers with lot of expandable enums. Having it in runtime is useful in general.

JonathanGiles:
I'm fine with ExpandableStringEnum (assuming it's purpose is clearly documented), but by default we should try to avoid using it unless it makes sense? In the case of HttpMethod, this should be a plain enum?

anuchandy:
Agree, let's make HttpMethod native enum

anuchandy:
Closing this.

PR #572, opened 2019-03-08
Making HttpMethod native enum

anuchandy:


anuchandy:
Addressing https://github.com/Azure/autorest-clientruntime-for-java/issues/552

PR #573, opened 2019-03-08
Fix tests based on httpbin changes and add proxy scheme support

daschult:
Resolves https://github.com/Azure/azure-storage-java/issues/445.
This PR fixes some stale tests and also adds proxy scheme support. Previously the scheme/protocol that was used to communicate with a proxy server was the same as the scheme for the final destination URL. There are some situations where a customer doesn't want that to be the case, though. In the referenced issue, the customer wants to use HTTP to communicate with their proxy server, but then they want the proxy server to use HTTPS to send the final request. This PR enables that.


anuchandy:
@daschult just checking, did you run the `RestProxyWithHttpProxyNettyTests` with this change?

daschult:
@anuchandy No, I didn't. I didn't even know about those tests. @jianghaolu is doing some proxy testing for me since he already has a proxy setup in place.

anuchandy:
@daschult based on our discussion today morning, I'm closing this.

PR #579, opened 2019-03-12
Updates to RestResponse API and various implementation details

JonathanGiles:
This is a draft PR. Please review and give feedback, and I will improve as necessary. The thing I dislike most is the RestResponseBase 'customHeaders' API. Feedback is welcome.


JonathanGiles:
Tagging @anuchandy and @conniey for code review please!

client-runtime/src/main/java/com/microsoft/rest/v3/RestProxy.java
```diff
@@ -406,93 +364,61 @@ private static Exception instantiateUnexpectedException(Class<? extends RestExce
...
+        Object[] args = new Object[ctor.getParameterCount()];
+        args[0] = httpRequest;
+        args[1] = responseStatusCode;
+        args[2] = responseHeaders;
+
+        if (args.length > 3) {
+            args[3] = bodyAsObject;
+        }
+        if (args.length > 4) {
+            args[4] = response.decodedHeaders().block();;
```
conniey: nit: double semi-colons.

client-runtime/src/main/java/com/microsoft/rest/v3/RestProxy.java
```diff
@@ -406,93 +364,61 @@ private static Exception instantiateUnexpectedException(Class<? extends RestExce
...
+        }
+
+        // we assume there is a single constructor for each type, and just take the first
+        Constructor<? extends RestResponse<?>> ctor = (Constructor<? extends RestResponse<?>>) cls.getConstructors()[0];
+
+        // create the constructor args array
+        Object[] args = new Object[ctor.getParameterCount()];
+        args[0] = httpRequest;
+        args[1] = responseStatusCode;
+        args[2] = responseHeaders;
```
conniey: Maybe add a check that the `ctor.getParameterCount()` be at least 3, and throw an exception if not or a Debug.assert (i don't know if those exist in Java).

conniey:
Looks good and makes the code easier to reason about.

client-runtime/src/main/java/com/microsoft/rest/v3/RestProxy.java
anuchandy: SDK implementors or others using our RestProxy can extends from one of the runtime/common response types and specify the extended type as the return type of their proxy method which we instantiate using reflection. In this case it would be good to have a check for -

1. At least one constructor and
2. with at least 3 parameters and
3. Even more defensive check for exact types of first 3 params (HttpRequest.class, Integer.TYPE, HttpHeaders.class)

client-runtime/src/main/java/com/microsoft/rest/v3/http/rest/SimpleRestResponse.java
```diff
@@ -0,0 +1,67 @@
...
+import com.microsoft.rest.v3.http.HttpRequest;
+
+import java.util.Map;
+
+/**
+ * REST response with a strongly-typed content specified.
+ *
+ * @param <T> The deserialized type of the response content.
+ */
+public class SimpleRestResponse<T> implements RestResponse<T> {
```
anuchandy: Was initially thinking why we need `SimpleRestResponse` class type given we support returning `RestResponse` interface type. Thinking more on this, it make sense to have `SimpleRestResponse`, e.g. SDKs returning their own response type with no custom/deserialized headers can extend from this type instead of RestResponseBase. This way those response types don't endup with  Void custom headers.

anuchandy:
@conniey I'm unable to directly respond to your comment https://github.com/Azure/autorest-clientruntime-for-java/pull/579#commitcomment-32704736, so adding it here.

yes, consumers of SDK do have access to raw headers and possibly could deserialize their own as you pointed. the thing is, swagger has this concept of strongly typed header model, there are many services using this feature. This is so common use case in azure, thats why we abstracted it to runtime level. We wanted to avoid customers deserializing these types themselves. 

e.g. From swagger we generate HeaderType like this https://github.com/Azure/azure-storage-java/blob/12d9633c87a7531f56223783ce944a2da7c8872d/src/main/java/com/microsoft/azure/storage/blob/models/BlobDeleteHeaders.java#L23. Then in runtime, once we receive the response raw headers, the matching properties of the HeaderType (match via json annotation) gets populated from the raw headers.

anuchandy:
@JonathanGiles could you run `mvn checkstyle:check` from root? it seems there are some unused imports.

JonathanGiles:
Thanks for the feedback, I'll do another iteration today.

client-runtime/src/main/java/com/microsoft/rest/v3/RestProxy.java
JonathanGiles: Good ideas - I'll improve this today.

client-runtime/src/main/java/com/microsoft/rest/v3/http/rest/SimpleRestResponse.java
JonathanGiles: Yes, I've always been on the fence about this, but I think it makes sense still.

conniey:
@anuchandy Thanks for that explanation and link! Its usecase makes more sense to me now. 👍

client-runtime/src/main/java/com/microsoft/rest/v3/RestProxy.java
```diff
@@ -406,93 +364,61 @@ private static Exception instantiateUnexpectedException(Class<? extends RestExce
...
+        final HttpHeaders responseHeaders = httpResponse.headers();
 
-            return asyncResult;
-        } catch (ReflectiveOperationException e) {
-            throw new RuntimeException(e);
+        // determine the type of response class. If the type is the 'RestResponse' interface, we will use the
+        // 'RestResponseBase' class instead.
+        Class<? extends RestResponse<?>> cls = (Class<? extends RestResponse<?>>) TypeUtil.getRawClass(entityType);
+        if (cls.equals(RestResponse.class)) {
+            cls = (Class<? extends RestResponse<?>>)(Object) RestResponseBase.class;
```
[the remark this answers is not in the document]
JonathanGiles: Tests added.

client-runtime/src/main/java/com/microsoft/rest/v3/RestProxy.java
```diff
@@ -406,94 +366,82 @@ private static Exception instantiateUnexpectedException(Class<? extends RestExce
...
+                throw reactor.core.Exceptions.propagate(e);
             }
-
-            return asyncResult;
-        } catch (ReflectiveOperationException e) {
-            throw new RuntimeException(e);
         }
+
+        // error
+        throw new RuntimeException("Cannot find suitable constructor for class " + cls);
```
conniey: Is it possible to add this as the `default:` case? I think it makes it easier to follow rather than falling out of the switch case. IIRC, spotbugs has a check where it would give you a warning if there was no default case, too.

conniey: Never mind. I see that its in a for loop. :) Ignore this.

JonathanGiles: I considered that (because this is my preference too), however, there should never be a default case - we've filtered to only get constructors with between 3 and 5 params.

PR #583, opened 2019-03-13
Refactoring of the client-runtime module

JonathanGiles:
Created an `implementation` package for non-public API, and moved a lot of classes into there.
Also, improved the javadocs for the classes in the annotations package.

API review can be done via the published [JavaDoc artifacts](https://azuresdkartifacts.blob.core.windows.net/azure-sdk-for-java/apidocs/azure-common/index.html).

// cc @anuchandy @conniey


client-runtime/src/main/java/com/microsoft/rest/v3/annotations/GET.java
```diff
@@ -14,19 +14,20 @@
 /**
  * HTTP GET method annotation describing the parametrized relative path to a REST endpoint for resource retrieval.
  *
- * The required value can be either a relative path or an absolute path. When it's
- * an absolute path, it must start with a protocol or a parametrized segment.
- * (Otherwise the parse cannot tell if it's absolute or relative)
+ * <p>The required value can be either a relative path or an absolute path. When it's an absolute path, it must start
+ * with a protocol or a parametrized segment (otherwise the parse cannot tell if it's absolute or relative).</p>
```
conniey: nit: spelling. parameterized.

client-runtime/src/main/java/com/microsoft/rest/v3/annotations/UnexpectedResponseExceptionType.java
```diff
@@ -6,7 +6,7 @@
 
 package com.microsoft.rest.v3.annotations;
 
-import com.microsoft.rest.v3.RestException;
+import com.microsoft.rest.v3.http.rest.RestException;
```
conniey: This existed before, but I was wondering why this it is called "UnexpectedResponseExceptionType" rather than just "UnexpectedResponseException" 🤔

client-runtime/src/main/java/com/microsoft/rest/v3/annotations/GET.java
JonathanGiles: Thanks, will fix. To be fair (due to your comment below thanking me), I didn't write any of the docs, I just geeked out and formatted them so they don't appear in the javadoc output in a massive blob of text.

client-runtime/src/main/java/com/microsoft/rest/v3/annotations/UnexpectedResponseExceptionType.java
JonathanGiles: I will look into this - it seems like a reasonable suggestion.

client-runtime/src/main/java/com/microsoft/rest/v3/annotations/GET.java
conniey: Ahh. Gotcha. :) I have a new appreciation for javadocs because of this.. it made it really easy to reason about the public surface area.

JonathanGiles: Yeah, you don't want to review the javadoc output of this repo before this PR is merged....it is just one big paragraph of text. Also, just committed a fix to all typos of parameterised....I mean, parameterized :-)

client-runtime/src/main/java/com/microsoft/rest/v3/annotations/UnexpectedResponseExceptionType.java
JonathanGiles: I could go either way on this. Shorter is normally better, and I don't believe the 'Type' adds any particular value to the name. Keen to get @anuchandy's feedback.

JonathanGiles: Let's leave as is for now.

JonathanGiles:
The Environment type was not touched, I left it where it was.

JonathanGiles:
I think this PR can be merged then?

anuchandy:
Merging.

PR #586, opened 2019-03-13
Adapting to new package structure.

anuchandy:


anuchandy:
Few things I wanted to highlight: 

1. The main reason for the name `client-runtime` was to communicate that it has nothing specific to azure. Now we’ve the name `azure-common` so non-azure SDKs builders will be bit reluctant or confused when using it. 
2. Moved `AzureEnvironment` to `azure-common`. Built-in Environment types exposes endpoints for both Azure ARM and Data plane services.
3. The `azure-common-auth` module used to take dependency on `azure-common-mgmt` now it depends on `azure-common` (This move required doing #2)

Please share thoughts on this.

JonathanGiles:
For easier review, people may consider browsing Anu's branch here that contains the new structure: https://github.com/anuchandy/autorest-clientruntime-for-java/tree/package-restruct

azure-common-auth/pom.xml
```diff
@@ -48,8 +49,8 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.microsoft.azure.v3</groupId>
-      <artifactId>azure-client-runtime</artifactId>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-common</artifactId>
       <version>3.0.0-SNAPSHOT</version>
```
JonathanGiles: We should consider rolling back the versions to 1.0.0 prior to release. Thoughts?

pom.xml
```diff
@@ -6,13 +6,13 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <groupId>com.microsoft.azure.v3</groupId>
-  <artifactId>autorest-clientruntime-for-java</artifactId>
+  <groupId>com.azure</groupId>
+  <artifactId>azure-common-for-java</artifactId>
```
JonathanGiles: I would propose we call this `azure-common-parent`

anuchandy: will call it `azure-common-parent`

azure-common-auth/pom.xml
anuchandy: sure, 3.x.x is not valid any more. May be we can start with `0.1.0` to indicate that it's in preview?

pom.xml
JonathanGiles: Let's just go with 1.0.0-SNAPSHOT. If we then want to do preview, we will [version it according to the Java spec](https://azuresdkspecs.z5.web.core.windows.net/JavaSpec.html#sec-preview-releases) (so, 1.0.0-alpha1, 1.0.0-beta1, or 1.0.0-rc1).

anuchandy: just pushed with 0.1.0, let me change it to 1.0.0-SNAPSHOT. Thanks for the reference.

azure-common-auth/pom.xml
anuchandy: Going with 1.0.0-SNAPSHOT as discussed below.

PR #590, opened 2019-03-18
Adding base class for RestPagedResponse<T>

conniey:
1. Adds ability for, given a RestProxy interface below, could deserialize this into a `RestPagedResponse<T>` without any user implemented classes. Users can implement their own `Page<T>` for `@ReturnValueWireType`, if the returned response does not match the generally known pagination patterns (ie. "value" with "nextPageLink"/"nextLink".)
```java
interface Service2 {
    @POST("anything/json")
    @ExpectedResponses({200})
    @ReturnValueWireType(KeyValuePage.class)
    RestPagedResponse<KeyValue> getPage(@BodyParam(ContentType.APPLICATION_JSON) Page<KeyValue> values);

    @POST("anything/json")
    @ExpectedResponses({200})
    @ReturnValueWireType(Page.class)
    Mono<RestPagedResponse<KeyValue>> getPageAsync(@BodyParam(ContentType.APPLICATION_JSON) Page<KeyValue> values);
}
```

2. In MockHttpClient, adds an "echo.org" so it returns the request body in the response. (Httpbin.org and the other ones added headers and other values that made it hard to test.)


azure-common/src/main/java/com/azure/common/http/rest/RestPagedResponseBase.java
```diff
@@ -0,0 +1,77 @@
...
+    private final String nextLink;
+
+    public RestPagedResponseBase(HttpRequest request, int statusCode, HttpHeaders headers, T body, H deserializedHeaders) {
+        this.request = request;
+        this.statusCode = statusCode;
+        this.headers = headers;
+        this.items = body.items();
+        this.nextLink = body.nextLink();
+        this.deserializedHeaders = deserializedHeaders;
+    }
```
JonathanGiles: Is this constructor necessary? I don't think I see it being used?

conniey: Ahh. This one is used in the reflection code when you figure out when to use 3 - 5 parameters: https://github.com/Azure/autorest-clientruntime-for-java/blob/a305b748094370d7fdf719deae3956db3e632343/azure-common/src/main/java/com/azure/common/implementation/RestProxy.java#L438-L439

JonathanGiles: How does the nextLink get set in that case?

conniey: I added a constraint that the class T needs to implement Page, which has nextLink(). And the object is deserialised already when it calls that constructor.

PR #592, opened 2019-03-21
Adding ClientConfiguration, ServiceClientBuilder

### LINKED ISSUE #588: [v3] Adding ServiceClient methods/ctors
@anuchandy @JonathanGiles Any suggestions would be helpful. :)

I've been trying to use Azure App Config with our new ServiceClient, and I've had to do a bit of weirdness because ServiceClient takes in a HttpPipeline for its constructor.

Problems that cropped up:
1. Customers calling AzConfigClient may to want to pass in their own HttpClient (ie. proxy)
1. Customers using AzConfigClient may want to add additional HttpPipelinePolicies in conjunction with our default ones. (ie. additional logging or headers)

As a result, I added in [PipelineOptions.java](https://github.com/conniey/azure-sdk-for-java/blob/fixes/applicationconfig/client/src/main/java/com/azure/applicationconfig/PipelineOptions.java).

Code:
[AzConfigClient.java](https://github.com/conniey/azure-sdk-for-java/blob/fixes/applicationconfig/client/src/main/java/com/azure/applicationconfig/AzConfigClient.java#L52-L56)
```java
class AzConfigClient {
    public AzConfigClient(String connectionString, PipelineOptions pipelineOptions) {
        super(createPipeline(new ApplicationConfigCredentials(connectionString), pipelineOptions));

        this.service = RestProxy.create(ApplicationConfigService.class, this);
        // Oddness is here because I had to create the creds twice since the super class required the
        // HttpPipeline already.
        this.baseUri = new ApplicationConfigCredentials(connectionString).baseUri();
    }
}

class PipelineOptions {
    public HttpClient client() {}
    public Iterable<HttpPipelinePolicy> additionalPolicies() {}
}
```

Would it be easier to add additional constructors/methods to ServiceClient? My proposal would be:

```java
public abstract class ServiceClient {
    protected ServiceClient() {
        this(HttpClient.createDefault());
    }

    /**
     * Creates a ServiceClient with the given HttpClient and additional policies.
     * @param client HttpClient to use for service calls
     * @param additionalPolicies Any additional policies that should be added. These are executed after the default policies.
     */
    protected ServiceClient(HttpClient client, HttpPipelinePolicy... additionalPolicies) { }
    
    /**
     * Gets the default pipeline policies for this service client to execute.
     *
     * Example:
     * Azure Application Configuration requires: Custom CredentialPolicy, RetryPolicy, UserAgentPolicy, RequestIdPolicy
     */
    protected abstract List<HttpPipelinePolicy> getPipelinePolicies();
}
```

conniey:
This adds a builder pattern to our ServiceClients.

Builder methods are static and we want consumers to be able to create a default `ServiceClient` easily, so we require the necessary arguments and always give them a valid ClientConfiguration.

Looking at our SDKs and others, the missing required parameter appears to be some way to authenticate with the service.  Unfortunately, we have no way to infer what their credentials may be from the environment. If they try to perform `.build()` on a ServiceClient, if the current configuration is not enough, it will throw an InvalidConfigurationException.

## Questions
* There is `AsyncServiceClientCredentials` and `ServiceClientCredentials`. Since we are moving to Mono and async, would it be better to use AsyncServiceClientCredentials by default and remove ServiceClientCredentials?

## Example ServiceClient
```java
public class ConfigurationClient extends ServiceClient {
    public static ConfigurationClientBuilder builder) {
        ClientConfiguration configuration = new ClientConfiguration();
        // implementers can add additional default values, etc. 
        return new ConfigurationClientBuilder(configuration);
    }

   static class ConfigurationClientBuilder {
        private ConfigurationClientBuilder(ClientConfiguration configuration) {
            this.configuration = configuration;
        }

        public static ConfigurationClient build() {
            // Closest to API goes first, closest to wire goes last.
            final List<HttpPipelinePolicy> policies = new ArrayList<>();
            // any additional policies to be added from configuration...
            HttpPipeline pipeline = config.getHttpClient() == null
                    ? new HttpPipeline(policies)
                    : new HttpPipeline(config.getHttpClient(), policies);
            return new ConfigurationClient(pipeline);
        }
    }
}
```

## Example SDK User
SDK users would have a flow like:
```java
public static void main() {
    ConfigurationClient client = ConfigurationClient.builder()
                                                    .credentials(new ConfigurationServiceCredentials("connectionString"))
                                                    .userAgent("MyAgent")
                                                    .addPolicy(PolicyThatIsAddedAtEnd)
                                                    .build();
}
```
Fixes #588 .


azure-common/src/main/java/com/azure/common/configuration/ClientConfiguration.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * Sets the credentials to use when authenticating HTTP requests.
+     *
+     * @param credentials The credentials to use for authenticating HTTP requests.
+     * @return The updated ClientConfiguration object.
+     */
+    public ClientConfiguration credentials(AsyncServiceClientCredentials credentials) {
+        Objects.requireNonNull(credentials);
```
JonathanGiles: Whenever you have an unchecked exception, please add an `@throws` JavaDoc statement that explains what triggers that exception to arise, e.g. `@throws NullPointerException if {@code credentials} is {@code null}`

azure-common/src/main/java/com/azure/common/configuration/InvalidConfigurationException.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+public class InvalidConfigurationException extends RuntimeException {
+    /**
+     * Creates an instance with the given message.
+     * @param message String describing why the configuration is invalid.
+     */
+    public InvalidConfigurationException(String message) {
+        super(message);
+    }
+}
```
JonathanGiles: If you aren't going to add any additional context information into this class, I would remove it and use IllegalStateException instead. Soon I will be speccing up the list of common exceptions we will use across the SDK, and ISE is one of them.

conniey: Gotcha. I'll remove this.

PR #593, opened 2019-03-21
Update AsyncServiceClientCredentials to take HttpRequest

conniey:
There are auth schemes that require more information than just the service url. Example is the HMAC-SHA256 that the Azure Application Configuration team requires for their service. This requires us to construct a string from HttpHeaders in addition to the service url.


conniey:
Opened PR in https://github.com/Azure/azure-sdk-for-java/pull/3200

Issue #569, opened 2019-03-05
Running azure-storage-java code sample warns of illegal restricted access operation


mattvryan:
I'm using [Azure Storage SDK for Java](https://github.com/Azure/azure-storage-java) and trying out the code examples.  I tried a very simple example based as closely as I could on the code sample in [their README.md](https://github.com/Azure/azure-storage-java/blob/master/README.md#code-sample). When I run my code, it creates the container, uploads the blob, verifies it, and then deletes the blob and container as expected. But warning messages are printed to the console about an illegal restricted access operation in the SDK code:

> WARNING: An illegal reflective access operation has occurred
> WARNING: Illegal reflective access by com.microsoft.rest.v2.Validator to field java.util.HashMap.serialVersionUID
> WARNING: Please consider reporting this to the maintainers of com.microsoft.rest.v2.Validator
> WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
> WARNING: All illegal access operations will be denied in a future release

I'm reluctant to rely on this SDK in production with this warning present; are there plans to address this?

mattvryan:
This was originally submitted to the Azure Storage SDK for Java project via [this issue](https://github.com/Azure/azure-storage-java/issues/447); at their request I've refiled it here.

lgaudez:
Any plans to address this ? 

  ```
  SharedKeyCredentials credentials = new SharedKeyCredentials(accountName, accountKey);
  HttpPipeline pipeline = StorageURL.createPipeline(credentials); // Execute this line pops the warnings
```

**Comment by Bruno-Brant** [comment]
I'm having the same issue, using the downstream package mentioned above (azure-storage-blob, v10.5.0). I traced the issue down to a line on the file Validator.java (here shows up on line 112)

    field.setAccessible(true);

This causes the warning. I'm not that well-versed in Java, but I believe that the reason this is happening is that (for some reason) the JVM is not allowing a third-party package to change class member access (nor I think that should be the way of going about it, if you care about it 😊).

I don't know exactly what is the end result of this WARNING (the code seems to work), but this kind of warning in production might cause problems (even if of a more ...political... nature).

**Comment by lawrencegripper** [comment]
I'm also seeing this error, any movement on a fix?

```
WARNING: An illegal reflective access operation has occurred
[...]

PR #597, opened 2019-03-28
 Validator checks just accessible fields (#569)

lpiedade:
JVM 9+ raises the message "

> All illegal access operations will be denied in a future release

because of this statement the Validator needs to try to set true to field access mode, it avoids exceptions in environments running JVM with option --illegal-access=deny.
**The Method trySetAccessible is Java 9 API Level**.


lpiedade:
The issue #569 just happens running on JRE 9+ and this bugfix needs JDK 9, but the pipeline is for Android environment.

PR #599, opened 2019-04-05
Support java.net.Authenticator for proxy by default

jianghaolu:


jianghaolu:
Closing this as after discussion, we decided not to make this the default.

Anyone seeking this feature should simply call `.withProxyAuthenticator(new JavaNetAuthenticator())`.

PR #636, opened 2019-09-28
Add support of LRO for postOrDelete.

yaohaizh:
Need this support to update the latest resource swagger spec for the whatif operation?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,9 @@ hs_err_pid*
 .idea
 *.iml
 .DS_Store
+
+# Eclipse/VSCode Java files
+.settings/
+.project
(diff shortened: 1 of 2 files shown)


jianghaolu:
Can we delete the other overload? https://github.com/Azure/autorest-clientruntime-for-java/blob/4207431c0f801ba30bbc681007981b5d1d4c1d79/azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java#L492

yaohaizh:
@jianghaolu this overload method is used by other generated methods thus we can't delete that version. Thanks. Also, could you help make this fix on the next release? Thanks.

PR #639, opened 2019-10-14
fix client id inside AzureCliSubscription

xccc-msft:


xccc-msft:
The previous fix doesn't make sense when it comes getToken() inside AzureCliSubscription. Hence, we have to make the fix again.

jianghaolu:
There's a small checkstyle error: src/main/java/com/microsoft/azure/credentials/AzureCliSubscription.java:[47,11] (whitespace) WhitespaceAround: 'if' is not followed by whitespace.

PR #643, opened 2019-11-06
Fix security warning.

yaohaizh:
Fix security alerts on the github repo main page.

### DIFF
--- a/build-tools/src/main/resources/checkstyle.xml
+++ b/build-tools/src/main/resources/checkstyle.xml
@@ -31,11 +31,7 @@
 -->
 
 <module name="Checker">
-    <module name="SuppressWithNearbyCommentFilter">
-        <property name="commentFormat" value="CHECKSTYLE IGNORE (\w+) FOR NEXT (\d+) LINE"/>
-        <property name="checkFormat" value="$1"/>
-        <property name="influenceFormat" value="$2"/>
-    </module>
+
     <module name="SuppressionFilter">
         <property name="file" value="${samedir}/suppressions.xml"/>
         <property name="optional" value="true"/>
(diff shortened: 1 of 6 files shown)


build-tools/src/main/resources/checkstyle.xml
```diff
@@ -84,11 +80,17 @@
...
+            <property name="commentFormat" value="CHECKSTYLE IGNORE (\w+) FOR NEXT (\d+) LINE"/>
+            <property name="checkFormat" value="$1"/>
+            <property name="influenceFormat" value="$2"/>
+        </module>
         <!-- Checks for Javadoc comments.                     -->
         <!-- See http://checkstyle.sf.net/config_javadoc.html -->
 
         <module name="JavadocMethod">
             <property name="scope" value="public"/>
+            <property name="allowMissingParamTags" value="true"/>
```
jianghaolu: Why is this needed?

build-tools/src/main/resources/checkstyle.xml
```diff
@@ -173,8 +175,6 @@
         <!-- See http://checkstyle.sf.net/config_modifiers.html -->
         <module name="ModifierOrder"/>
         <module name="RedundantModifier"/>
-        <module name="FileContentsHolder"/>
```
jianghaolu: Why is this removed? From the docs it seems this is needed if used in conjunction with `SuppressWithNearbyCommentFilter`: https://checkstyle.sourceforge.io/version/6.18/config_filters.html#SuppressWithNearbyCommentFilter

>  Usage: This filter only works in conjunction with a FileContentsHolder, since that check makes the suppression comments in the Java files available. A configuration that includes this filter must configure FileContentsHolder as a child module of TreeWalker.

yaohaizh: According to this release note: https://checkstyle.sourceforge.io/releasenotes.html#Release_8.2, this is not necessary again after version 8.2. Here we use 8.24 to make it consist with the latest setup on the azure-sdk-for-java-repo.

build-tools/src/main/resources/checkstyle.xml
yaohaizh: There are too many warnings about violation of this rule. Given this, I suppress this temporarily. If the security work is not urgent, I would like to fix the issue on generator and then on the fluent code base.

PR #655, opened 2020-02-19
Update dependencies.

conniey:
Updates guava and jackson dependencies.

Related to: https://github.com/Azure/azure-sdk-for-java/pull/8247

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>20.0</version>
+        <version>24.1.1-jre</version>
       </dependency>
       <dependency>
         <groupId>com.squareup.retrofit2</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -56,7 +56,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>20.0</version>
+        <version>24.1.1-jre</version>
```
srnagar: Doesn't this also need a version update for this pom itself?

srnagar: Nevermind, this is a SNAPSHOT. So, don't need to bump up the version.

jianghaolu:
@yaohaizh Would you please review the impact for management plane?

yaohaizh:
@jianghaolu when will we publish 1.7.2 for this update?

pom.xml
jianghaolu: The versions are automatically updated with maven release plugin.

PR #665, opened 2020-04-17
Added flag for keeping track of when value is set in ServiceFuture

vcolin7:
Code in this PR is based on discussion on [this issue](https://github.com/azure/azure-sdk-for-java/issues/6553) filed for `azure-keyvault-extensions` (Track 1 SDK).

### DIFF
--- a/client-runtime/src/main/java/com/microsoft/rest/ServiceFuture.java
+++ b/client-runtime/src/main/java/com/microsoft/rest/ServiceFuture.java
@@ -201,7 +201,8 @@ public class ServiceFuture<T> extends AbstractFuture<T> {
      * @return true if successfully reported; false otherwise.
      */
     public boolean success(T result) {
-        return valueSet = set(result);
+        valueSet = set(result);
+        return valueSet;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


vcolin7:
@anuchandy @jianghaolu would you mind giving this a quick look? Thanks :)

vinayak-kuchekar:
Hello, We are actually waiting for this fix to be available.
Please try to publish the changes ASAP.

Thanks & Regards,
Vinayak Kuchekar

client-runtime/src/main/java/com/microsoft/rest/ServiceFuture.java
```diff
@@ -43,7 +44,7 @@ protected ServiceFuture() {
             .subscribe(new Action1<ServiceResponse<T>>() {
                 @Override
                 public void call(ServiceResponse<T> t) {
-                    serviceFuture.set(t.body());
+                    serviceFuture.valueSet = serviceFuture.set(t.body());
```
anuchandy, line 47: should we make the `valueSet` flag a volatile? In case `isCancelled` gets called from a thread other than the subscribe thread, in such case, there is a possibility that `isCancelled` calling thread won't see the new flag value immediately.

PR #676, opened 2020-07-05
Add timeout to ApplicationTokenCredentials

jianghaolu:


jianghaolu:
Zip containing the jar with the fix:
[azure-client-authentication-1.7.6-SNAPSHOT.jar.zip](https://github.com/Azure/autorest-clientruntime-for-java/files/4875755/azure-client-authentication-1.7.6-SNAPSHOT.jar.zip)

azure-client-authentication/src/main/java/com/microsoft/azure/credentials/ApplicationTokenCredentials.java
```diff
@@ -39,15 +41,19 @@
  */
 public class ApplicationTokenCredentials extends AzureTokenCredentials {
     /** A mapping from resource endpoint to its cached access token. */
-    private Map<String, AuthenticationResult> tokens;
+    private final ConcurrentHashMap<String, AuthenticationResult> tokens;
+    /** A mapping from resource endpoint to its current authentication futures. */
```
haogang, line 45: `futures` -> `locks`?

azure-client-authentication/src/main/java/com/microsoft/azure/credentials/ApplicationTokenCredentials.java
```diff
@@ -132,17 +140,54 @@ String clientCertificatePassword() {
...
+            }
         }
-        tokens.put(resource, authenticationResult);
         return authenticationResult.getAccessToken();
     }
 
-    private AuthenticationResult acquireAccessToken(String resource) throws IOException {
+    private Future<AuthenticationResult> acquireAccessToken(String resource) throws IOException {
         String authorityUrl = this.environment().activeDirectoryEndpoint() + this.domain();
         ExecutorService executor = Executors.newSingleThreadExecutor();
```
haogang, line 192: @jianghaolu now that this function can return before `context.acquireToken` finishes, and it shuts down the executor before returning (L221), can we guarantee that `context.acquireToken` does not submit new tasks to the executor asynchronously? In that case, the async tasks might be rejected when `acquireAccessToken` returns early.

It seems that the interface doesn't guarantee this cannot happen, and the correctness of the code depends on the implementation detail of `AuthenticationContext.acquireToken`. Can we at least add a test to make sure we don't accidentally break the assumption in the future?

jianghaolu: Thanks for catching this! It does seem like the executor can be shut down before the Future gets executed. Let me fix this and add a unit test.
I tested the current code with AAD but a smoke test is probably insufficient for real-world workloads.

PR #680, opened 2020-08-26
support okhttp protocols

weidongxu-microsoft:
For https://github.com/Azure/azure-libraries-for-java/issues/1197

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>24.1.1-jre</version>
+        <version>20.0</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -56,7 +56,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>24.1.1-jre</version>
+        <version>20.0</version>
```
weidongxu-microsoft, line 59: 20.0 is later version supports Java 7
https://github.com/google/guava/wiki/ReleaseHistory

weidongxu-microsoft:
@jianghaolu 
I have no write access to the repo hence I cannot add anybody as reviewer.

I've downgraded guava. Let me know if it is OK or not.

PR #681, opened 2020-08-28
Allow public access to subscriptions in AzureCliCredentials

jianghaolu:


weidongxu-microsoft (approved):

azure-client-authentication/src/main/java/com/microsoft/azure/credentials/AzureCliCredentials.java
```diff
@@ -118,6 +119,13 @@ public AzureEnvironment environment() {
         return subscriptions.get(defaultSubscriptionId()).environment();
     }
 
+    /**
+     * @return the subscriptions the current account in Azure CLI has access to
+     */
+    public Map<String, AzureCliSubscription> subscriptions() {
+        return subscriptions;
```
weidongxu-microsoft, line 126: Should we do an unmodifiable one?

jianghaolu: yeah good idea

weidongxu-microsoft (approved):

azure-client-authentication/src/main/java/com/microsoft/azure/credentials/AzureCliCredentials.java
```diff
@@ -118,6 +119,13 @@ public AzureEnvironment environment() {
         return subscriptions.get(defaultSubscriptionId()).environment();
     }
 
+    /**
+     * @return the subscriptions the current account in Azure CLI has access to
+     */
+    public Map<String, AzureCliSubscription> subscriptions() {
```
wangmingliang-ms, line 125: I think you should also make `class AzureCliSubscription` public, it's not 'visible' from outside packages at current.

PR #695, opened 2020-12-22
handle exceptions during lro polling

xccc-msft:
Before: The error will be ignored if it's not `CloudException`.
After: It would throw the error if it is `CloudException` or last retry attempt is completed.


xccc-msft:
@jianghaolu @weidongxu-microsoft @ChenTanyi

weidongxu-microsoft (approved):

ChenTanyi (approved):

jianghaolu (approved):

PR #696, opened 2021-01-04
Backport flattening deserializer to track 1

jianghaolu:
Types referenced in flattened types are not handled properly: https://github.com/Azure/autorest.java/issues/873


weidongxu-microsoft (approved):
I've ran playback test with track1 mgmt with this new update. All passes.

PR #699, opened 2021-01-07
run the delay on Schedulers.immediate

weidongxu-microsoft:
Original `delaySubscription` without specifying a specific scheduler seems causing deadlock in replay mode on certain Linux version (ubuntu 18.04).

Now put it to `Schedulers.immediate()` similar to existing code, e.g. https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java#L258-L261

Replay sets to run on `Schedulers.trampoline()` (i.e., single thread), which might be the cause. But still not figured out why OS matters (there seems nothing for OS to schedule if that is already a single thread).

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/AzureClient.java
@@ -133,7 +133,7 @@ public final class AzureClient extends AzureServiceClient {
                             return Observable.just(pollingState);
                         } else {
                             // initial delay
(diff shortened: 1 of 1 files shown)


ChenTanyi (commented):
I think it may need to change in two place? according to https://github.com/Azure/autorest-clientruntime-for-java/pull/694/files

weidongxu-microsoft:
@ChenTanyi Thanks. Fixed.

ChenTanyi (approved):

PR #703, opened 2021-02-26
Add policy evaluation details related classes

lacikaaa:
`PolicyViolationErrorInfo` is missing `@JsonIgnoreProperties(ignoreUnknown = true)` annotation and `ObjectMapper` wasn't configured for ignoring unknown fields.
This caused an exception when Azure returned `evaluationDetails` in it's response, making impossible to handle the error gracefully.

In this commit:
- Configured `ObjectMapper` with `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` to avoid issues when a new field is introduced on Azure side
- created `EvaluationDetails` and subsequent classes to parse the response
- added `toString` implementation for related and newly created classes for easier logging
- added unit test to verify the parsing

### DIFF
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/EvaluationDetails.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/EvaluationDetails.java
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure;
+
+import java.util.List;
+
+public class EvaluationDetails {
+
+    private List<ExpressionEvaluationDetails> evaluatedExpressions;
+
+    public List<ExpressionEvaluationDetails> getEvaluatedExpressions() {
+        return evaluatedExpressions;
+    }
+
+    @Override
+    public String toString() {
+        return "EvaluationDetails{"
+                + "evaluatedExpressions=" + evaluatedExpressions
+                + '}';
+    }
+}
--- a/azure-client-runtime/src/main/java/com/microsoft/azure/ExpressionEvaluationDetails.java
+++ b/azure-client-runtime/src/main/java/com/microsoft/azure/ExpressionEvaluationDetails.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure;
+
+import java.util.List;
+
+public class ExpressionEvaluationDetails {
+
+    private String expression;
+
+    private String expressionValue;
+
+    private String operator;
+
+    private String path;
+
+    private String result;
+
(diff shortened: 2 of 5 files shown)


lacikaaa:
@weidongxu-microsoft could you take a look? thanks

weidongxu-microsoft:
Add @jianghaolu for review.

I am not familiar with this part. But I think fail on known properties should be already turned off?
https://github.com/Azure/autorest-clientruntime-for-java/blob/master/client-runtime/src/main/java/com/microsoft/rest/serializer/JacksonAdapter.java#L151

OK maybe the `ObjectMapper` in `CloudErrorDeserializer` is on another configure.

lacikaaa:
> OK maybe the ObjectMapper in CloudErrorDeserializer is on another configure. Do we want to turn FAIL_ON_UNKNOWN_PROPERTIES off there as well?

Shall I make this change? I'm happy to do it if it fits more to the current solution.

weidongxu-microsoft:
> > OK maybe the ObjectMapper in CloudErrorDeserializer is on another configure. Do we want to turn FAIL_ON_UNKNOWN_PROPERTIES off there as well?
> 
> Shall I make this change? I'm happy to do it if it fits more to the current solution.

Let us wait a bit for Jianghao's input.

weidongxu-microsoft (approved):

weidongxu-microsoft:
@lacikaaa

BTW, which lib does you see this problem?

Asking because we had new generation SDK, but it is a bit more simplistic on policy violation error details. Considering whether we should improve it.
https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core-management/src/main/java/com/azure/core/management/exception/AdditionalInfo.java

lacikaaa:
we are using multiple which depend on this:
- https://mvnrepository.com/artifact/com.microsoft.azure/azure
- https://mvnrepository.com/artifact/com.microsoft.azure/azure-mgmt-datalake-store
- https://mvnrepository.com/artifact/com.microsoft.azure/azure-mgmt-sql

I think it would be a bigger effort on our side to switch to the new SDK even if it already supports everything we need.

thanks for the review!

weidongxu-microsoft:
@lacikaaa 

Sure, got it. Not asking you to migrate. Just want to make sure what is supported in current SDK will be supported in new SDK. So when people really need to migrate, they will find it easier.

lacikaaa:
How do you feel about adding `toString` implementation to these classes? It would make logging easier, but I haven't seen it in the existing ones.

We have seen these policy validations a few times. I think it's pretty helpful if the reason can be extracted.

jianghaolu:
It's missing here: https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-runtime/src/main/java/com/microsoft/azure/PolicyViolation.java#L37. The best fix is to add 
```java
.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```

lacikaaa:
@jianghaolu I have added the config and removed the annotations.
also added `toString` implementation.
I have updated the commit message to reflect the current state
Could you take a look? thanks

jianghaolu (approved):
LGTM

PR #720, opened 2024-04-20
feat: bump up for java version to 17

opensourcevk:
updated java version to 17 (LTS)

testing proof: 
all projects are compiling with Java 17
image


weidongxu-microsoft:
The project here is for compatibility of Java 7. And this lib likely won't have another release.

